Issue #7, opened 2017-12-05
Link to GitHub organisation for each Google Code-in organisation 


jayvdb:
Link to the GitHub org for each Google Code-in organisation, wherever possible, ensuring that only highly probable links are made.

Use a GitHub logo for the link.

jayvdb:
It seems you've just enabled the issue triaging. I'm just scraping all issues from your repository and will give you some more information about this in a few minutes or so.

Because of the rate limit we can't scrape all information (including all comments and authors) right now - our system is already set up to scrape this in the next days over which the predictions will become more precise every day.

**Sit tight!**

jayvdb:
[GitMate.io](https://gitmate.io) thinks the label prediction failed because there are not enough labels. Three labels must have been set on at least three issues each.

The contributor prediction failed because there are not enough contributors. Three contributors must have authored comments on at least three issues each.

PR #15, opened 2017-12-05
Find GitHub org for each GCI org

andrewda:
This will add the ability to find the GitHub organization of a GCI
org by looking in different links and the description, and then
searching GitHub if the previous search was unsuccessful.

Closes https://github.com/coala/gci-leaders/issues/7

## Logic

_(first successful step will be considered the organization's GitHub org)_

- ~~Check if overridden~~
- RegEx check for `https://github.com/(orgname)` or `https://gitter.im/(orgname)` in one of org URLs 
- RegEx check for `https://github.com/(orgname)` in org description
- GitHub search for org name (must have >10 score)
- `null`

## Accuracy

`100%`

All organizations' GitHub accounts were successfully found. 10 are found in the GCI org, and the other 15 are found through a GitHub API call.

## Screenshot

screen shot 2017-12-04 at 11 58 01 pm

### DIFF
--- a/generate.js
+++ b/generate.js
@@ -47,8 +47,15 @@ const template = `
     <div class="orgs">
       {{#orgs}}
         <div class="org">
-          <h3><a href="https://codein.withgoogle.com/organizations/{{slug}}">{{name}}</a></h3>
-          <p>Task Completed: {{completed_task_instance_count}}</p>
+          <h3>
+            <a href="https://codein.withgoogle.com/organizations/{{slug}}">{{name}}</a>
+            <p>Task Completed: {{completed_task_instance_count}}</p>
+            {{#github}}
+            <a href="https://github.com/{{github}}">
+              <img src="https://assets-cdn.github.com/images/modules/logos_page/GitHub-Mark.png" height="14" />
+            </a>
+            {{/github}}
+          </h3>
           <ul>
             {{#leaders}}
               <li>{{display_name}}</li>
--- a/scrap.js
+++ b/scrap.js
@@ -1,6 +1,10 @@
 const fetch = require('node-fetch')
 const fs = require('fs')
 
+const GITHUB_API_BASE = 'https://api.github.com'
+
+const MIN_SEARCH_SCORE = 10
+
 async function fetchOrgs() {
   const res = await fetch(
     'https://codein.withgoogle.com/api/program/2017/organization/?status=2'
@@ -17,20 +21,82 @@ async function fetchLeaders(id) {
   return leaders
 }
 
-async function fetchOrgsWithLeaders() {
+async function searchGitHubOrgs(query) {
+  const token = process.env.GITHUB_TOKEN
+  const res = await fetch(
+    `${GITHUB_API_BASE}/search/users?q=${query}%20type:org`,
+    {
+      headers: token ? { Authorization: `token ${token}` } : {}
+    }
+  )
+  const { items } = await res.json()
+  return items || []
+}
+
+async function findOrganization({
+  name,
+  description,
+  mailing_list,
+  website_url,
+  irc_channel,
+  blog_url,
+  guide_to_working_url,
+}) {
+  const ghPattern = /(?:https?:\/\/)?(?:github\.com|gitter\.im)\/([a-zA-Z0-9-]+)/i
+  const websites = [
+    mailing_list,
+    website_url,
+    irc_channel,
+    blog_url,
+    guide_to_working_url,
+  ]
+
+  const orgFromWebsites = websites
+    .map(website => (ghPattern.exec(website) || [])[1])
+    .find(org => org)
+
+  if (orgFromWebsites) {
+    return orgFromWebsites
+  }
+
+  const orgFromDescription = (ghPattern.exec(description) || [])[1]
+
+  if (orgFromDescription) {
+    return orgFromDescription
+  }
+
+  console.warn(
+    `Could not find GitHub org for ${name}. Resorting to GitHub API hit.`
+  )
+
+  const removePattern = /the|project|\([a-zA-Z]+\)/gi
+  const searchQuery = name.replace(removePattern, '').trim()
+  const searchResults = await searchGitHubOrgs(searchQuery)
+
+  if (searchResults.length > 0 && searchResults[0].score > MIN_SEARCH_SCORE) {
+    return searchResults[0].login
+  }
+
+  return null
+}
+
+async function fetchOrgsWithData() {
   const orgs = await fetchOrgs()
   const fetchingLeaders = orgs.map(org => fetchLeaders(org.id))
+  const fetchingGitHub = orgs.map(org => findOrganization(org))
   const orgLeaders = await Promise.all(fetchingLeaders)
+  const orgGitHub = await Promise.all(fetchingGitHub)
 
   return orgs.map((org, index) =>
     Object.assign(org, {
       leaders: orgLeaders[index],
+      github: orgGitHub[index],
     })
   )
 }
 
 ;(async () => {
-  const data = await fetchOrgsWithLeaders()
+  const data = await fetchOrgsWithData()
   
   // sort data by completed_task_instance_count
   data.sort((a, b) => 


andrewda:
A `GITHUB_TOKEN ` env variable must be added to Netlify for build to pass.

scrap.js
```diff
@@ -17,19 +19,82 @@ async function fetchLeaders(id) {
   return leaders
 }
 
-async function fetchOrgsWithLeaders() {
+async function searchGitHubOrgs(query) {
+  const token = process.env.GH_TOKEN
```
blazeu: [`.travis.yml`](https://github.com/coala/gci-leaders/blob/master/.travis.yml#L10) uses `GITHUB_TOKEN` var.
So ideally this should use the same var.

Why do we even need this? API rate limit?

scrap.js
```diff
@@ -17,19 +19,82 @@ async function fetchLeaders(id) {
...
+    mailing_list,
+    website_url,
+    irc_channel,
+    blog_url,
+    guide_to_working_url,
+  ]
+
+  const orgFromWebsites = websites
+    .map(website => (githubPattern.exec(website) || [])[1])
+    .filter(org => org)
```
blazeu: What's this filter?

scrap.js
```diff
@@ -17,19 +19,82 @@ async function fetchLeaders(id) {
...
+async function searchGitHubOrgs(query) {
+  const token = process.env.GH_TOKEN
+  const res = await fetch(
+    `https://api.github.com/search/users?q=${query}%20type:org`,
+    {
+      headers: token ? { Authorization: `token ${token}` } : {}
+    }
+  )
+  const resp = await res.json()
+  const { items } = resp
```
blazeu: `const { items } = await res.json()`
is 1 line shorter.

blazeu (dismissed):

scrap.js
andrewda: Yea, the API does not like us making a bunch of requests at once without a token. I'll change it to `GITHUB_TOKEN` now!

scrap.js
andrewda: Yep, sorry, some debug code I forgot to get rid of!

scrap.js
andrewda: The map will return some `undefined` elements. This filter gets rid of them because `undefined` is falsey.

andrewda:
One of Travis or Netlify is failing. I'm guessing it's whichever one runs last as the first one can successfully make the requests without the token and without rate limiting.

scrap.js
blazeu: Does the resulting array only contains 1 truthy element?
If so you can use `.find` .

andrewda: It may contain more than one element, but we only care about the first one, so you're right, `.find` is probably the better option!

scrap.js
```diff
@@ -17,23 +19,85 @@ async function fetchLeaders(id) {
...
   )
 }
 
 ;(async () => {
-  const data = await fetchOrgsWithLeaders()
-  
+  const data = await fetchOrgsWithData()
+
   // sort data by completed_task_instance_count
-  data.sort((a, b) =>
```
blazeu: Dang, seems like we need to add coala ASAP.

Can you not mess with this ugly whitespacing problem for now?

andrewda: Sure, sorry about that, atom automatically fixed it for me :P

scrap.js
```diff
@@ -17,21 +19,83 @@ async function fetchLeaders(id) {
...
     Object.assign(org, {
       leaders: orgLeaders[index],
+      github: orgGitHub[index],
     })
   )
 }
 
 ;(async () => {
-  const data = await fetchOrgsWithLeaders()
-
```
blazeu: This line too, please.
You can check with git diff before pushing to make sure that everything is tidy.

github-overrides.json
```diff
@@ -0,0 +1,4 @@
+{
+  "CloudCV": "Cloud-CV",
```
jayvdb: See https://codein.withgoogle.com/api/program/current/organization/ ; this value is already there.

github-overrides.json
```diff
@@ -0,0 +1,4 @@
...
+  "Liquid Galaxy project": "LiquidGalaxyLAB"
```
jayvdb: This value is also in the JSON, and can be extracted as part of the fallback strategy to find the correct exact match.

scrap.js
```diff
@@ -17,20 +19,82 @@ async function fetchLeaders(id) {
   return leaders
 }
 
-async function fetchOrgsWithLeaders() {
+async function searchGitHubOrgs(query) {
+  const token = process.env.GITHUB_TOKEN
+  const res = await fetch(
+    `https://api.github.com/search/users?q=${query}%20type:org`,
+    {
+      headers: token ? { Authorization: `token ${token}` } : {}
```
jayvdb: I dont mind if a fork's build without a token envvar results in some orgs missing the link.
But the site should work without env vars.
i.e. degrading gracefully.

Orgs have the option of improving their GCI profile data to make their GitHub org easily findable, if they want to.
And forkers can add the env var to travis.

We will also be able to have a few orgs resolved using GitLab, when that enhancement is done, and then infer the GitHub from the GitLab.

jayvdb (changes requested):

scrap.js
```diff
@@ -17,20 +17,78 @@ async function fetchLeaders(id) {
...
+async function findOrganization({
+  name,
+  description,
+  mailing_list,
+  website_url,
+  irc_channel,
+  blog_url,
+  guide_to_working_url,
+}) {
+  const githubPattern = /(?:https?:\/\/)?(?:github\.com|gitter\.im)\/([a-zA-Z0-9-]+)/i
```
jayvdb: This will be a long line that another task is going to have to fix.
And you will loose ownership (blame) of these lines when they fix it.

scrap.js
```diff
@@ -17,20 +17,78 @@ async function fetchLeaders(id) {
...
+    return orgFromWebsites
+  }
+
+  const orgFromDescription = (githubPattern.exec(description) || [])[1]
+
+  if (orgFromDescription) {
+    return orgFromDescription
+  }
+
+  const removePattern = /the|project|\([a-zA-Z]+\)/gi
```
jayvdb: Add stderr here to indicate above algorithms failed and resorting to a API hit.

generate.js
```diff
@@ -47,8 +47,15 @@ const template = `
...
       {{#orgs}}
         <div class="org">
-          <h3><a href="https://codein.withgoogle.com/organizations/{{slug}}">{{name}}</a></h3>
-          <p>Task Completed: {{completed_task_instance_count}}</p>
+          <h3>
+            <a href="https://codein.withgoogle.com/organizations/{{slug}}">{{name}}</a>
+            <p>Task Completed: {{completed_task_instance_count}}</p>
+            {{#github}}
+            <a href="https://github.com/{{github}}">
+              <img src="https://image.flaticon.com/icons/png/512/25/25231.png" height="14" />
```
blazeu: Is this icon "safe" to use without putting a credit?

scrap.js
```diff
@@ -17,20 +17,78 @@ async function fetchLeaders(id) {
...
+  const searchQuery = name.replace(removePattern, '').trim()
+  const searchResults = await searchGitHubOrgs(searchQuery)
+
+  if (searchResults.length > 0 && searchResults[0].score > 10) {
```
jayvdb: 10 is a constant.  it should be treated as a constant.

scrap.js
```diff
@@ -17,20 +17,78 @@ async function fetchLeaders(id) {
   return leaders
 }
 
-async function fetchOrgsWithLeaders() {
+async function searchGitHubOrgs(query) {
+  const token = process.env.GITHUB_TOKEN
+  const res = await fetch(
+    `https://api.github.com/search/users?q=${query}%20type:org`,
```
jayvdb: base url is a constant that may be needed again in future enhancements.

jayvdb (changes requested):

blazeu:
Just a heads-up, this repo will use prettier.io code formatter via ESLintBear.

Ideally you should start formatting it now (using `prettier` directly)

andrewda:
Yep, used prettier on this code already!

generate.js
andrewda: That's a good point. IANAL but I'd assume so because it's GitHub's logo.

andrewda: I should be able to use the official GitHub icon, right?

https://assets-cdn.github.com/images/modules/logos_page/GitHub-Mark.png

jayvdb (approved):

jayvdb:
ack https://github.com/coala/gci-leaders/pull/15/commits/78d0eaa12835fb15e1cecbb0e48a9b84fc97c4b3

Issue #34, opened 2017-12-06
ESLint should extend from eslint:recommended and eslint-config-prettier rules


blazeu:
https://github.com/coala/gci-leaders/pull/20 only adds prettier which focuses on formatting rules.

`eslint:recommended` adds some real linter rules. e.g. unused variable.
[`prettier`](https://github.com/prettier/eslint-config-prettier) rules is used to remove any formatting rules that conflicts with prettier.

jayvdb:
[GitMate.io](https://gitmate.io) thinks the contributor prediction failed because there are not enough contributors. Three contributors must have authored comments on at least three issues each.

andrewda:
My bad for not adding these in #20. Assuming everything passes CI, #37 should add this.

PR #37, opened 2017-12-06
Extend ESLint rules to use recommended

andrewda:
This extends ESLint rules to use the `eslint:recommended` settings as
well as extending a `prettier` config to ensure that nothing breaks.

Closes https://github.com/coala/gci-leaders/issues/34

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,14 @@
 ---
+env:
+  node: true
+  browser: true
+  es6: true
 parserOptions:
   ecmaVersion: 2017
 plugins:
+  - prettier
+extends:
+  - eslint:recommended
   - prettier
 rules:
   prettier/prettier:
@@ -11,3 +18,4 @@ rules:
       trailingComma: es5
       bracketSpacing: true
       arrowParens: avoid
+  no-console: 0
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,7 @@
     "acorn": {
       "version": "5.2.1",
       "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.2.1.tgz",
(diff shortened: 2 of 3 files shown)


.eslintrc
```diff
@@ -1,8 +1,15 @@
 ---
+env:
+  node: true
```
blazeu, line 3: Use `shared-node-browser`
as we'll be linting browser specific js too.

blazeu (changes requested):

.eslintrc
andrewda: Using `shared-node-browser` gives me errors:

```
.../gci-leaders/generate.js
  1:12  error  'require' is not defined  no-undef
  2:18  error  'require' is not defined  no-undef
  3:13  error  'require' is not defined  no-undef
  4:14  error  'require' is not defined  no-undef

.../gci-leaders/scrap.js
   1:15  error  'require' is not defined  no-undef
   2:12  error  'require' is not defined  no-undef
  25:17  error  'process' is not defined  no-undef
```

Is adding on `browser: true` beneath equivalent? That seems to work fine.

blazeu: Yea, I usually use `browser` and `node` 

But `shared-node-browser` is listed on https://eslint.org/docs/user-guide/configuring#specifying-environments :thinking:

andrewda: Weird, ok. I can look into that more in a while, but for now adding both seems fine 👍

.eslintrc
```diff
@@ -1,8 +1,16 @@
...
+  browser: true
```
blazeu, line 4: Add `es6` as well?
Maybe that will make `Promise` available ..

blazeu: `parserOptions` is for setting the syntax, that does not include globals

andrewda: Yep, adding `es6` to the env made Promise available!

blazeu (approved):

blazeu:
ack https://github.com/coala/gci-leaders/pull/37/commits/9a14e0b433252e563a4097ce1cb7162217b5346a

Issue #29, opened 2017-12-06
Display only 1 column on mobile


blazeu:
https://github.com/coala/gci-leaders/issues/21 should be done first.

jayvdb:
[GitMate.io](https://gitmate.io) thinks the contributor prediction failed because there are not enough contributors. Three contributors must have authored comments on at least three issues each.

PR #39, opened 2017-12-06
static/css/main.css: Responsive column

tioirawan:
Make column responsive in different viewport width

Closes https://github.com/coala/gci-leaders/issues/29

### DIFF
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -31,3 +31,21 @@ ul {
   margin: 0;
   padding-left: 1em;
 }
+
+@media only screen and (max-width : 450px) {
+  .org {
+    flex: 0 0 100%;
+  }
+}
+
+@media only screen and (min-width : 450px) and (max-width : 600px) {
+  .org {
+    flex: 0 0 50%;
+  }
+}
+
+@media only screen and (min-width : 600px) and (max-width : 750px) {
+  .org {
+    flex: 0 0 33.333333%;
+  }
+}


static/css/main.css
```diff
@@ -31,3 +31,21 @@ ul {
   margin: 0;
   padding-left: 1em;
 }
+
+@media only screen and (max-device-width : 450px) {
```
blazeu: Use `max-width` instead.
`max-device-width` does not respond to browser resize.

blazeu (changes requested):

static/css/main.css
```diff
@@ -31,3 +31,21 @@ ul {
...
+  }
+}
+
+@media only screen and (min-width : 450px) and (max-width : 600px) {
+  .org {
+    flex: 50%;
+  }
+}
+
+@media only screen and (min-width : 600px) and (max-width : 750px) {
```
blazeu: What happens if you remove the `min-width`, I think it should work fine?

tioirawan: no, it doesn't work

static/css/main.css
```diff
@@ -31,3 +31,21 @@ ul {
   margin: 0;
   padding-left: 1em;
 }
+
+@media only screen and (max-width : 450px) {
+  .org {
+    flex: 100%;
```
blazeu: Should be `flex-basis` or `flex: 0 0 <size>` :P
Setting it like this would be like doing `flex: 1 1 100%` which could make the column expand.

Although it does not affect the visual now, it might do in the future.
Change all the occurence of `flex: ..` in all breakpoints.

blazeu (changes requested):

static/css/main.css
```diff
@@ -31,3 +31,21 @@ ul {
...
+
+@media only screen and (min-width : 450px) and (max-width : 600px) {
+  .org {
+    flex: 50%;
+  }
+}
+
+@media only screen and (min-width : 600px) and (max-width : 750px) {
+  .org {
+    flex: 33%;
```
blazeu: This doesnt add up to 100% ..

tioirawan: what? this is for 3 column

blazeu: Yeah, but 33 * 3 is not 100

Check out bootstrap grid
https://github.com/twbs/bootstrap/blob/v4-dev/dist/css/bootstrap-grid.css#L128-L131

blazeu (approved):

blazeu:
ack https://github.com/coala/gci-leaders/pull/39/commits/5bcc5e12801e7593a50fa56a94066c2a1d3bd00b

blazeu:
Submit your GCI task

Issue #17, opened 2017-12-05
Place orgs with no listed leaders in separate section 


jayvdb:
The orgs without any leaders are visually unappealing amongst the other orgs.

They should be a separate list, at the bottom of the page, ordered by number of tasks completed.

jayvdb:
[GitMate.io](https://gitmate.io) thinks the label prediction failed because there are not enough labels. Three labels must have been set on at least three issues each.

The contributor prediction failed because there are not enough contributors. Three contributors must have authored comments on at least three issues each.

PR #43, opened 2017-12-07
Separate orgs with no listed leaders

tioirawan:
This will filter all orgs div with 'none' class
and put them at the bottom of the page

Closes https://github.com/coala/gci-leaders/issues/17

### DIFF
--- a/generate.js
+++ b/generate.js
@@ -12,11 +12,18 @@ ncp('static', 'out/static', err => {
   }
 })
 
+const noLeader = []
+const withLeader = orgs.filter(org => {
+  if (!org.leaders.length) noLeader.push(org)
+  return org.leaders.length
+})
+
 fs.writeFileSync(
   'out/index.html',
   Mustache.render(fs.readFileSync('templates/main.html').toString(), {
-    orgs,
+    withLeader,
     datetime,
     rootURL,
+    noLeader,
   })
 )
--- a/templates/main.html
+++ b/templates/main.html
@@ -22,7 +22,7 @@
       The leading participants for each organization are listed randomly. 
     </i>
     <div class="orgs">
-      {{#orgs}}
+      {{#withLeader}}
         <div class="org">
           <h3>
             <a href="https://codein.withgoogle.com/organizations/{{slug}}">
@@ -47,8 +47,32 @@
             {{/leaders}}
           </ul>
         </div>
-      {{/orgs}}
+      {{/withLeader}}
     </div>
+    {{#noLeader.length}}
+      <h3>Organization(s) with no listed leaders</h3>
+      <hr>
+      <div class="orgs">
+        {{#noLeader}}
+          <div class="org">
+            <h3>
+              <a href="https://codein.withgoogle.com/organizations/{{slug}}">
+                {{name}}
+              </a>
+              <p>Task Completed: {{completed_task_instance_count}}</p>
+              {{#github}}
+              <a href="https://github.com/{{github}}">
+                <img
+                  src="https://assets-cdn.github.com/images/modules/logos_page/GitHub-Mark.png"
+                  height="18"
+                />
+              </a>
+              {{/github}}
+            </h3>
+          </div>
+        {{/noLeader}}
+      </div>
+    {{/noLeader.length}}
     <footer>
       <small>Google Code-in and the Google Code-in logo are trademarks of Google Inc.</small>
     </footer>


templates/main.html
```diff
@@ -49,6 +49,9 @@ <h3>
         </div>
       {{/orgs}}
     </div>
+    <h4>Organization with no listed leader</h4>
+    <hr>
+    <div class="orgs" id="no-leader"></div>
```
blazeu: We can generate this on build time,
so that it works without client side js.

blazeu (changes requested):

templates/main.html
blazeu: Also do check if there's no organization without leaders,
don't show the heading/line "organization with no listed leaders"

tioirawan:
how to fix error with coala? i type coala and it just checking without error

tioirawan:
screenshot from 2017-12-07 17-11-42

blazeu:
You might need run this 2 line inside the project folder

```
export PATH=$PATH:./node_modules/.bin
export NODE_PATH=$NODE_PATH:./node_modules
```

To let coala use the local eslint.

generate.js
```diff
@@ -12,11 +12,25 @@ ncp('static', 'out/static', err => {
...
 })
 
+let i = orgs.length
+let noLeader = []
+while (i--) {
+  if (!orgs[i].leaders.length) {
+    noLeader.push(orgs[i])
+    orgs.splice(i, 1)
+  }
+}
```
blazeu: You could use `.filter`
`.push` can be done inside the callback.

generate.js
```diff
@@ -12,11 +12,25 @@ ncp('static', 'out/static', err => {
...
+
+noLeader.sort(
+  (a, b) => b.completed_task_instance_count - a.completed_task_instance_count
+)
```
blazeu: Sorting should be done first.

generate.js
```diff
@@ -12,11 +12,25 @@ ncp('static', 'out/static', err => {
   }
 })
 
+let i = orgs.length
+let noLeader = []
```
blazeu: Use `const` , we dont need to reassign this variable.

templates/main.html
```diff
@@ -49,6 +49,30 @@ <h3>
...
+              <img
+                src="https://assets-cdn.github.com/images/modules/logos_page/GitHub-Mark.png"
+                height="18"
+              />
+            </a>
+            {{/github}}
+          </h3>
+        </div>
+      {{/noLeader}}
+    </div>
```
blazeu: Put the check of no leaders to here.

blazeu (changes requested):

generate.js
blazeu: ... but if you're using filter you need to reassign this variable, so let it be

generate.js
Indmind: I try to do this
```es6
const noLeader = orgs.filter((org, index) => {
  if (!org.leaders.length){
    orgs.splice(index, 1)
    return true
  }
  return false
})
```
but if i splice orgs array, ..., should I make temp array? any suggestion

tioirawan: if i use filter and splice orgs, it will not working

blazeu: `filter` will remove the element from an array if the callback returns false.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

blazeu: But it's returning a new array so you need to do 
```js
orgs = orgs.filter(..
```

tioirawan:
or like this
```es6
let orgs = require('./out/data.json')

const noLeader = []
orgs = orgs.filter(org => {
  if (!org.leaders.length) {
    noLeader.push(org)
    return false
  }
  return true
})
```

generate.js
```diff
@@ -12,11 +12,22 @@ ncp('static', 'out/static', err => {
   }
 })
 
+// slice(0) is to make new array, so it will not modified even though orgs is modified
+const orgsTemp = orgs.slice(0)
+const noLeader = orgsTemp.filter((org, index) => {
+  if (!org.leaders.length) {
+    orgs.splice(orgs.findIndex(o => o.id === orgsTemp[index].id), 1)
```
blazeu: I bet it is simpler to just run `filter` on orgs directly and use `noLeader.push` here.

generate.js
```diff
@@ -12,11 +12,22 @@ ncp('static', 'out/static', err => {
...
+    return true
+  }
+  return false
```
blazeu: `return org.leaders.length` will remove the need of 2 `return`.

blazeu (changes requested):

generate.js
tioirawan: so, i need to change orgs from const to let?

blazeu: You can copy it.
Resulting 2 new array, withLeader & noLeader.

blazeu: We might have another use for the original orgs variable

templates/main.html
```diff
@@ -47,8 +47,32 @@ <h3>
             {{/leaders}}
           </ul>
         </div>
-      {{/orgs}}
+      {{/withLeader}}
     </div>
+    {{#noLeader.length}}
+      <h4>Organization with no listed leader</h4>
```
blazeu: Organizations

Also `<h2`> probably looks better, let's test it!

tioirawan: doesn't it look too big?

blazeu: Yeah :/
Maybe try h3?

also "leaders"

templates/main.html
```diff
@@ -47,8 +47,32 @@ <h3>
...
+      <h2>Organization(s) with no listed leader</h2>
+      <hr>
+      <div class="orgs">
+        {{#noLeader}}
+          <div class="org">
+            <h3>
+              <a href="https://codein.withgoogle.com/organizations/{{slug}}">
+                {{name}}
+              </a>
+              <p>Task Completed: {{completed_task_instance_count}}</p>
```
HizkiFW: This should be changed to `Tasks Completed` since it's more than one. The same for line 31.

templates/main.html
```diff
@@ -47,8 +47,32 @@ <h3>
...
+      <h3>Organization(s) with no listed leaders</h3>
+      <hr>
+      <div class="orgs">
+        {{#noLeader}}
+          <div class="org">
+            <h3>
+              <a href="https://codein.withgoogle.com/organizations/{{slug}}">
+                {{name}}
+              </a>
+              <p>Tasks Completed: {{completed_task_instance_count}}</p>
```
blazeu: Noooooo don't change it ...

You can make the issue, however.

blazeu (approved):

Issue #58, opened 2017-12-10
Create deploy script


blazeu:
for Travis.

jayvdb:
Note that this one does not need to use ssh deploy; as this build process doesnt contain private information.  Up to you, but ssh deploy makes it a bit more difficulty for the average forker.
Maybe try to have 'ssh deploy if configured, or fallback to normal overwrite GitHub Pages deploy'

PR #59, opened 2017-12-10
Create SSH deploy script

blazeu:
With fallback to Travis deployment.

Closes https://github.com/coala/gci-leaders/issues/58


blazeu:
Note that I used SSH keys because it's more secure than token.

.ci/deploy.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+set -e # Exit with nonzero exit code if anything fails
```
naveenkumarsangi: Also set two other flags 
- `-x ` to show the commands being run.
- `-o pipefail` to fail if any of the commands in a pipefail.

blazeu: Actually don't want `-x` because it'll output some private var?

.travis.yml
```diff
@@ -1,6 +1,11 @@
 language: python
 python: 3.6
 
+env:
+  global:
+    - ENCRYPTION_LABEL: "b7c548322a41"
```
jayvdb: As this is invalid on a fork, I think these should be in Travis environment variables. (also `COMMIT_*`).

.ci/deploy.sh
```diff
@@ -0,0 +1,75 @@
...
+git clone $REPO out
+cd out
+git checkout $TARGET_BRANCH || git checkout --orphan $TARGET_BRANCH
+cd ..
+
+# Clean out existing contents
+rm -rf out/*
+
+# Run our compile script
+npm run build
```
jayvdb: In my version, the build is not in the deploy script.  The deploy script should take a built version, and overlay it on the previous built version.

.ci/deploy.sh
```diff
@@ -0,0 +1,75 @@
...
+set -eo pipefail # Exit with nonzero exit code if anything fails
+
+SOURCE_BRANCH="master"
+TARGET_BRANCH="gh-pages"
+
+# Pull requests and commits to other branches shouldn't try to deploy, just build to verify
+if [ "$TRAVIS_PULL_REQUEST" != "false" -o "$TRAVIS_BRANCH" != "$SOURCE_BRANCH" ]; then
+    echo "Skipping deploy; just doing a build."
+    # Don't waste token on PR
+    unset GITHUB_TOKEN
```
jayvdb: the PR wont get an env var on Travis, or any other sane system (Netlify isnt sane/safe), so this is useless on Travis.

but more importantly, this is useless.  If the token is available here, a malicious person could modify this script to emit it to the log , or send it via email to themselves, or anything.

However this isnt such a large problem if the token in GITHUB_TOKEN has no scopes which allow greater access than the malicious person could obtain themselves.  The worst they can do is abuse the token so that GitHub cancels it, breaking the builds.

jayvdb (changes requested):

.ci/deploy.sh
blazeu: Yeah, didn't realize Travis remove all env var on PR.

The token has no scope, unless you want to do deployment using Travis (which in this case we don't).

.travis.yml
```diff
@@ -10,7 +10,7 @@ before_install:
 
 script:
   - coala --non-interactive -V
-  - npm run build
```
jayvdb: Can we please keep the build process in .travis.yml , not in a script called `deploy.sh` .

.ci/deploy.sh
```diff
@@ -0,0 +1,65 @@
...
+git checkout $TARGET_BRANCH || git checkout --orphan $TARGET_BRANCH
+mv .git ../out/.git
+cd ../out
+
+# Now let's go have some fun with the cloned repo
+git config user.name "Travis CI"
+git config user.email "$COMMIT_AUTHOR_EMAIL"
+
+# If there are no changes to the compiled out (e.g. this is a README update) then just bail.
+if git diff --quiet; then
```
blazeu: Actually this will do deploy all the time because of the "Last Updated" in the html ..

.travis.yml
```diff
@@ -19,3 +20,4 @@ deploy:
   local_dir: out
   "on":
     branch: master
+    condition: $TRAVIS_DEPLOY = true
```
jayvdb: maybe `condition: "$ENCRYPTION_LABEL" = ""` ?

jayvdb: (No idea if that syntax will work)

blazeu: >`condition`: deploy when a single bash condition evaluates to `true`

Maybe just put `-z "$ENCRYPTION_LABEL"` will work?

blazeu: ^ Yup, tested on my fork.

.ci/deploy.sh
```diff
@@ -0,0 +1,62 @@
...
+cd ../out
+
+# Now let's go have some fun with the cloned repo
+git config user.name "Travis CI"
+git config user.email "$COMMIT_AUTHOR_EMAIL"
+
+# If there are no real changes to the compiled out
+# e.g. this is a README update or build time update in html
+# then just bail.
+if git diff --quiet -- . ':(exclude)index.html'; then
```
blazeu: I'm still thinking about this, what if it changes the html layout ..

jayvdb: nah, dont bother with this exclude... just push it all.
The only improvement is keeping the `gh-pages` filestamps stable, so the json's last modification can be deduced

jayvdb:
ack https://github.com/coala/gci-leaders/pull/59/commits/79107ed4fc4a29bac371b0e5ba2db0a37f663893

jayvdb (approved):

.ci/deploy.sh
jayvdb: (approving; up to you what you do with this...)

blazeu: Yeah, I'll remove the exclude, it's overcomplicating.

.ci/deploy.sh
```diff
@@ -0,0 +1,61 @@
...
+SOURCE_BRANCH="master"
+TARGET_BRANCH="gh-pages"
+
+# Pull requests and commits to other branches shouldn't try to deploy, just build to verify
+if [ "$TRAVIS_PULL_REQUEST" != "false" -o "$TRAVIS_BRANCH" != "$SOURCE_BRANCH" ]; then
+    echo "Skipping deploy."
+    exit 0
+fi
+
+if [ -z "$ENCRYPTION_LABEL" ]; then
```
jayvdb, line 13: this makes `[ "$TRAVIS_PULL_REQUEST" != "false" .. ` irrelevant...?

PRs to any repo with `ENCRYPTION_LABEL` set:

- dont get that `ENCRYPTION_LABEL` set, or
- any manual addition of it in `.travis.yml` is malicious/stupid, but the decrypt part of the deploy below will fail.

.ci/deploy.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+set -eo pipefail # Exit with nonzero exit code if anything fails
+
+SOURCE_BRANCH="master"
+TARGET_BRANCH="gh-pages"
+
+# Pull requests and commits to other branches shouldn't try to deploy, just build to verify
+if [ "$TRAVIS_PULL_REQUEST" != "false" -o "$TRAVIS_BRANCH" != "$SOURCE_BRANCH" ]; then
```
jayvdb, line 8: Taking into account the `ENCRYPTION_LABEL` decision described below, the remaining case is

- pushes to a branch on https://github.com/coala/gci-leaders should not deploy to "gh-pages"
- pushes to a branch on https://github.com/jayvdb/gci-leaders should deploy to "gh-pages"

Thinking a bit broader, an additional case exists on the `community` project, because I want forks of `community` to be able to have their own network of forks. (without requiring conflicting changes) i.e.

- pushes to a branch on https://github.com/librehealthio/community should not deploy to "gh-pages"

The one way to achieve this is detecting that https://github.com/coala and https://github.com/librehealthio are orgs vs users, and give them different default behaviours.

A simpler way, is require that either orgs or users set a variable to choose different behaviour.  I think the easiest is the require that orgs do nothing, but a user needs to set `SOURCE_BRANCH="*"` , or something similar, to explicitly say I want to deploy any branch I push, overwriting stuff is OK.  (the `SOURCE_BRANCH="master"` above would need to change, obviously)

jayvdb:
ack https://github.com/coala/gci-leaders/pull/59/commits/962f85e1dbe9ba891217ffd71f2485b11b56e4ce

.ci/deploy.sh
blazeu: `ENCRYPTION_LABEL` is not something that must be kept in hidden env var, although I put it there.

Issue #32, opened 2017-12-06
Detect when display name is a GitHub username


jayvdb:
Split off from https://github.com/coala/gci-leaders/issues/8 , lets get the easy case done sooner than later.

jayvdb:
[GitMate.io](https://gitmate.io) thinks the contributor prediction failed because there are not enough contributors. Three contributors must have authored comments on at least three issues each.

PR #62, opened 2017-12-11
scrape.js: Detect if GCI name is GitHub username

andrewda:
This detects when a student's GCI display name is their GitHub
username. To ensure we don't get the wrong account, this will first
verify that the user has been active since the start of GCI. If they
have, we then check if the user has created an issue or PR to the
organization's GitHub org. If so, we consider them the correct user.

Closes https://github.com/coala/gci-leaders/issues/32

### DIFF
--- a/lib/scrape.js
+++ b/lib/scrape.js
@@ -4,6 +4,7 @@ const fs = require('fs')
 const json2yaml = require('json2yaml')
 
 const GITHUB_API_BASE = 'https://api.github.com'
+const GCI_API_BASE = 'https://codein.withgoogle.com/api'
 
 const MIN_SEARCH_SCORE = 10
 
@@ -24,22 +25,18 @@ const GITHUB_OPTIONS = {
 }
 
 async function fetchProgram() {
-  const res = await fetch('https://codein.withgoogle.com/api/program/2017/')
+  const res = await fetch(`${GCI_API_BASE}/program/2017/`)
   return await res.json()
 }
 
 async function fetchOrgs() {
-  const res = await fetch(
-    'https://codein.withgoogle.com/api/program/2017/organization/?status=2'
-  )
+  const res = await fetch(`${GCI_API_BASE}/program/2017/organization/?status=2`)
   const { results } = await res.json()
   return results
 }
 
 async function fetchLeaders(id) {
-  const res = await fetch(
-    `https://codein.withgoogle.com/api/program/current/organization/${id}`
-  )
+  const res = await fetch(`${GCI_API_BASE}/program/current/organization/${id}`)
   const { leaders } = await res.json()
   return leaders
 }
--- a/lib/scrape.js
+++ b/lib/scrape.js
@@ -17,6 +17,17 @@ const CHAT_IMAGES = {
   OTHER: 'static/images/chat.png',
 }
 
+const GITHUB_OPTIONS = {
+  headers: process.env.GITHUB_TOKEN
+    ? { Authorization: `token ${process.env.GITHUB_TOKEN}` }
+    : {},
+}
+
+async function fetchProgram() {
+  const res = await fetch('https://codein.withgoogle.com/api/program/2017/')
+  return await res.json()
+}
+
 async function fetchOrgs() {
   const res = await fetch(
     'https://codein.withgoogle.com/api/program/2017/organization/?status=2'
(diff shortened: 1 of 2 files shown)


blazeu:
Can we not use the search API again? :/

andrewda:
@blazeu Any idea how to do this _without_ the search API? We're actually making very few requests here with the current users (7 or 8 I believe). I played around with some other ideas, but if we don't use the search API, username like `@owen` might be mistakingly placed on a user whose display name is `Owen`.

blazeu:
Maybe check if the user has a forked repo from the org?

templates/main.html
```diff
@@ -48,7 +49,15 @@ <h3>
...
                 {{#leaders}}
-                  <li>{{display_name}}</li>
+                  <li>
+                    {{#github_account}}
+                    <a href="https://github.com/{{github_account}}">
+                    {{/github_account}}
+                    {{display_name}}
+                    {{#github_account}}
+                    </a>
+                    {{/github_account}}
```
blazeu: Indent properly for readability.

blazeu (changes requested):

lib/scrape.js
```diff
@@ -33,6 +35,30 @@ async function searchGitHubOrgs(query) {
...
+    }
+  )
+  const { items } = await res.json()
+  return items || []
+}
+
+async function getGitHubUser(user) {
+  const token = process.env.GITHUB_TOKEN
+  const res = await fetch(`${GITHUB_API_BASE}/users/${user}`, {
+    headers: token ? { Authorization: `token ${token}` } : {},
```
blazeu: Maybe it's time to put headers inside constant

lib/scrape.js
```diff
@@ -80,19 +106,56 @@ async function findOrganization({
   return null
 }
 
+async function findGitHubUser({ display_name }, org) {
+  if (org) {
```
blazeu: `if (!org) return`
to reduce nesting

lib/scrape.js
```diff
@@ -80,19 +106,56 @@ async function findOrganization({
...
+    const displayNameMatches = displayNamePattern.exec(display_name)
+    if (displayNameMatches) {
+      const user = await getGitHubUser(display_name)
+      const { competition_open_starts } = await fetchProgram()
+
+      if (user) {
+        const updatedTime = new Date(user.updated_at)
+        const openTime = new Date(competition_open_starts)
+
+        if (updatedTime.getTime() - openTime.getTime() > 0) {
```
blazeu: Same case with the 3 `if` here, inverting it will reduce nesting a lot

lib/scrape.js
```diff
@@ -91,6 +122,35 @@ async function findOrganization({
...
+  const updatedTime = new Date(user.updated_at)
+  const openTime = new Date(competition_open_starts)
+
+  if (updatedTime.getTime() - openTime.getTime() < 0) return
+
+  const nov = await getGitHubUserCommits(login, '2017-11-28', '2017-11-30')
+  const dec = await getGitHubUserCommits(login, '2017-12-01', '2017-12-31')
+  const jan = await getGitHubUserCommits(login, '2018-01-01', '2018-01-17')
+  const orgs = nov.concat(dec).concat(jan).map(repo => repo.split('/')[0])
+  if (orgs.indexOf(org) > -1) {
```
blazeu: orgs.includes ?

lib/scrape.js
```diff
@@ -91,6 +122,35 @@ async function findOrganization({
...
+
+  const updatedTime = new Date(user.updated_at)
+  const openTime = new Date(competition_open_starts)
+
+  if (updatedTime.getTime() - openTime.getTime() < 0) return
+
+  const nov = await getGitHubUserCommits(login, '2017-11-28', '2017-11-30')
+  const dec = await getGitHubUserCommits(login, '2017-12-01', '2017-12-31')
+  const jan = await getGitHubUserCommits(login, '2018-01-01', '2018-01-17')
+  const orgs = nov.concat(dec).concat(jan).map(repo => repo.split('/')[0])
```
blazeu: use array spread operator, I think node 8 supports it.

blazeu (approved):

lib/scrape.js
blazeu: Also why not directly do `find`

andrewda: Mainly just in case someone commits to the `random_org/coalas_are_cool` repository. Very unlikely to happen, but using the organization is a more clean approach IMO.

blazeu: I mean replacing map with find,
you could do split and string comparison there just fine :P

But whatever. This works fine.

jayvdb (approved):

jayvdb:
ack https://github.com/coala/gci-leaders/pull/62/commits/5641077802b6cc4802965adad1cd2dee882d8a34 https://github.com/coala/gci-leaders/pull/62/commits/ea7deddaed23a5f2fe02533e0dcb869da75ed7a0

Issue #8, opened 2017-12-05
Guess GitHub username of leaders


jayvdb:
Link to the GitHub user for the leaders, ensuring that only highly probable links are made.

At least confirm that a username is a member of the relevant GitHub organisation before linking to it.

This is a repeatable [GCI task](https://codein.withgoogle.com/dashboard/tasks/5179714090041344/).

Use "`Related to https://github.com/coala/gci-leaders/issues/8`" at the bottom of your commit message, so it doesnt close this issue.

jayvdb:
[GitMate.io](https://gitmate.io) thinks the label prediction failed because there are not enough labels. Three labels must have been set on at least three issues each.

The contributor prediction failed because there are not enough contributors. Three contributors must have authored comments on at least three issues each.

jayvdb:
Blocked on https://github.com/coala/gci-leaders/issues/7

andrewda:
Most GCI students are private members of the org (at least for coala) so this might be near impossible...

Edit: I did a manual check and if we're searching the organization for the user's name, currently the _only_ person it will work for is me. Everyone else is private or not in the company's organization.

jayvdb:
It is high difficulty, but I can assure you it is very possible to do this.

invisible-defects:
We can search user by name & surname, GitHub user search has such option.
In case we don't find someone, we can try to guess by training seq2seq model on `username : real name` pairs we've found so far. We know our guess is right if user has coala org in his profile, which is AFAIK visible for other users. Though it definitely won't work for participants, whose nicknames don't have any associations with their real names (e.g. me)

jayvdb:
This was a repeatable task.

PR #75, opened 2017-12-12
scrape.js: Search org for user

andrewda:
This searches the user's organization for their display name. If
the user is found, their username will be used.

Closes https://github.com/coala/gci-leaders/issues/8

### DIFF
--- a/.coafile
+++ b/.coafile
@@ -14,6 +14,7 @@ bears = LineLengthBear
 
 [all.links]
 bears = InvalidLinkBear
+link_ignore_regex = (github\.com|\{|\$)
 
 [js]
 files = lib/**/*.js, static/js/**/*.js
--- a/lib/scrape.js
+++ b/lib/scrape.js
@@ -2,8 +2,10 @@ const fetch = require('node-fetch')
 const chattie = require('chattie')
 const fs = require('fs')
 const json2yaml = require('json2yaml')
+const validUsername = require('valid-github-username')
 
 const GH_USER_BASE = 'https://github.com/users'
+const GH_ORG_BASE = 'https://github.com/orgs'
 const GH_API_BASE = 'https://api.github.com'
 const GCI_API_BASE = 'https://codein.withgoogle.com/api'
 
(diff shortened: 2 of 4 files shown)


lib/scrape.js
```diff
@@ -4,6 +4,7 @@ const fs = require('fs')
 const json2yaml = require('json2yaml')
 
 const GH_USER_BASE = 'https://github.com/users'
+const GH_ORG_BASE = 'https://github.com/orgs' // Ignore InvalidLinkBear
```
jayvdb: I think just add github.com to the ignore list in .coafile

lib/scrape.js
```diff
@@ -155,14 +167,17 @@ async function findOrganization({
 async function findGitHubUser({ display_name }, org) {
   if (!org) return
 
-  display_name = display_name.replace(/ /g, '')
+  const shortName = display_name.replace(/ /g, '')
```
jayvdb: spaces are not valid in github usernames, so this is a very basic "convert to valid username".  There was a npm package which checked legality of a name.  What we want is a smarter "convert name into most similar valid github probable username".  maybe a npm package already exists; if not, ... create one ;-)

jayvdb (changes requested):

lib/scrape.js
andrewda: Oh man, alright, I'm running out of package naming ideas ;)

.coafile
```diff
@@ -14,6 +14,7 @@ bears = LineLengthBear
 
 [all.links]
 bears = InvalidLinkBear
+link_ignore_regex = (github\.com|\{|\$)
```
blazeu, line 17: There's `link_ignore_list` that uses globbing instead
`link_ignore_list = https://github.com/**`

blazeu: ^ Optional.

jayvdb (approved):

jayvdb:
ack https://github.com/coala/gci-leaders/pull/75/commits/f2934af1278f4087732774f655afc273534d0d27

blazeu:
ack https://github.com/coala/gci-leaders/pull/75/commits/f2934af1278f4087732774f655afc273534d0d27

PR #76, opened 2017-12-12
.coafile Fixup

blazeu:
The current .coafile will only run section `all` only on .js file, we want it to run on all common files.

Also switch to stylelint because it has saner default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ out/*
 !out/.keep
 yarn.lock
 .DS_Store
+**.orig
--- a/.coafile
+++ b/.coafile
@@ -1,23 +1,25 @@
 [all]
-files = lib/*.js, static/js/*.js
+files = **.js, **.json, **.css, **.yml, **.html, .eslintrc
+ignore = node_modules/**, out/**, package-lock.json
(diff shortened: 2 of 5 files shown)


blazeu:
There's a weird bug on
https://travis-ci.org/coala/gci-leaders/builds/315169207#L770

Maybe because I was putting the same thing on `ignore` and `files` >> `**.html`

.gitignore
```diff
@@ -3,3 +3,4 @@ out/*
 !out/.keep
 yarn.lock
 .DS_Store
+**.orig
```
jayvdb, line 6: fyi, there is a flag to turn off creation of `.orig` files.

jayvdb:
ack https://github.com/coala/gci-leaders/pull/76/commits/c6740080f3dca02ed47de447c3335c0bba12b690 https://github.com/coala/gci-leaders/pull/76/commits/998ac110e8131f7f4b2a9afc1f6fbd0eb00d8920 https://github.com/coala/gci-leaders/pull/76/commits/fd1531838c9c0d65af74bde6ef6a6147498cea11

jayvdb (approved):

PR #79, opened 2017-12-12
Wikidata for organisations

### LINKED ISSUE #63: Add Wikidata & Wikipedia links for every org
Do searches on Wikidata, either using the `wbsearchentities` API (used by https://www.npmjs.com/package/wikidata-search) , or using [WDQS](https://www.mediawiki.org/wiki/Wikidata_query_service) (sparql)

nalinbhardwaj:
Adds links to wikipedia/wikidata pages if they exist. Prefers wikipedia entry over wikidata entry(since they are generally sparse), if org doesn't have an english wikipedia entry, it displays the wikidata URL. 

Uses [wikidata-sdk](https://www.npmjs.com/package/wikidata-sdk) to process URLs for Wikidata API. 

Preview:
screen shot 2017-12-13 at 2 41 41 am

Once i18n support is added, we could default to user's language, but left for now.

Closes #63.


lib/scrape.js
```diff
@@ -165,6 +166,47 @@ async function findOrganization({
   return null
 }
 
+async function findWiki(name) {
+  const APIurl = wdk.searchEntities({
+    search: name,
+    format: 'json',
+  })
+  const srch = await fetch(APIurl)
```
andrewda: `search`?

lib/scrape.js
```diff
@@ -165,6 +166,47 @@ async function findOrganization({
...
+    let sitelink = await sitelinks.json()
+    sitelink = sitelink['entities'][wikidataid]['sitelinks']
+    if (sitelink && sitelink['enwiki']) {
+      const wikipediaurl = wdk.getSitelinkUrl(sitelink['enwiki'])
+      return {
+        url: wikipediaurl,
+        type: 'wikipedia',
+      }
+    }
+    else {
```
andrewda: No newline before else

lib/scrape.js
```diff
@@ -165,6 +166,47 @@ async function findOrganization({
...
+      return {
+        url: 'https:' + wikidataurl,
+        type: 'wikidata',
+      }
+    }
+  }
+  else {
```
andrewda: No newline before else

lib/scrape.js
```diff
@@ -241,6 +285,10 @@ async function fetchOrgsWithData() {
         platform: chattie.CHAT[orgChats[index].type],
         image: CHAT_IMAGES[chattie.CHAT[orgChats[index].type]],
       },
+      wiki: {
+        url: orgWiki[index].url,
+        type: orgWiki[index].type,
+      }
```
andrewda: Add trailing comma

andrewda (changes requested):

lib/scrape.js
nalinbhardwaj: Was in the process of linting, done now.

lib/scrape.js
```diff
@@ -165,6 +166,45 @@ async function findOrganization({
...
+    sitelink = sitelink['entities'][wikidataid]['sitelinks']
+    if (sitelink && sitelink['enwiki']) {
+      const wikipediaurl = wdk.getSitelinkUrl(sitelink['enwiki'])
+      return {
+        url: wikipediaurl,
+        type: 'wikipedia',
+      }
+    } else {
+      return {
+        url: 'https:' + wikidataurl,
```
andrewda: This is probably not needed because `//` is actually totally valid for a URL -- it's protocol aware so it uses either `http` or `https`  depending on what the browser is using (though I could totally see wanting to do it this way out of consistency with other URLs).

lib/scrape.js
nalinbhardwaj: Sure, changing name.

lib/scrape.js
```diff
@@ -165,6 +166,45 @@ async function findOrganization({
...
+    const wikidataid = response.search[0].id
+    const entityAPIurl = wdk.getEntities({
+      ids: [wikidataid],
+      props: ['sitelinks'],
+      format: 'json',
+    })
+
+    const sitelinks = await fetch(entityAPIurl)
+    let sitelink = await sitelinks.json()
+    sitelink = sitelink['entities'][wikidataid]['sitelinks']
```
andrewda: In general I think doing `sitelink.entities[wikidataid].sitelinks` is more appropriate (and I think that's what we've been doing in other places), but either works. Is there an ESLint rule for this?

lib/scrape.js
```diff
@@ -165,6 +166,45 @@ async function findOrganization({
...
+    if (sitelink && sitelink['enwiki']) {
```
andrewda: I think here you should definitely do `sitelink.enwiki`

lib/scrape.js
nalinbhardwaj: Yep, but wikipedia/wikidata is guaranteed to provide HTTPS, not so much guaranteed for the gci-leaders website, I just think it's better to default to HTTPS?

andrewda: Sure that works!

andrewda: Maybe in that case `https:${wikidataurl}`?

andrewda:
Oh, one last thing, it might be better to use the Wikipedia and Wikimedia logos without text, because it's not readable anyways when the logo is so small. Using just the icon without text will help it fit in more.

lib/scrape.js
nalinbhardwaj: In future(with `i18n`) this should become language_code + 'wiki'. So that'd probably bring us back to this way...

andrewda: Ah, gotcha, that makes sense.

lib/scrape.js
nalinbhardwaj: I've decided to remove `https:`, since I just noticed netlify also does https, so there's no point to it then. Will put it back if anyone thinks it's a problem.

blazeu: FYI, netlify without custom domain doesn't have HTTPS enforcement, but this shouldn't be a problem at all.

lib/scrape.js
```diff
@@ -217,9 +257,11 @@ async function fetchOrgsWithData() {
   const fetchingLeaders = orgs.map(org => fetchLeaders(org.id))
   const fetchingGitHub = orgs.map(org => findOrganization(org))
   const fetchingChat = orgs.map(org => chattie(org.irc_channel))
+  const fetchingWiki = orgs.map(org => findWiki(org.name))
   const orgLeaders = await Promise.all(fetchingLeaders)
   const orgGitHub = await Promise.all(fetchingGitHub)
   const orgChats = await Promise.all(fetchingChat)
+  const orgWiki = await Promise.all(fetchingWiki)
```
blazeu, line 268: I wish we have native `Promise.map` to use it here :(

blazeu (approved):

blazeu:
I don't think you need to split the commit.

1 commit for 1 issue.

jayvdb (changes requested):
The task is to add Wikipedia & Wikidata links, not only one.

lib/scrape.js
```diff
@@ -165,6 +166,45 @@ async function findOrganization({
...
+
+    const entityAPIurl = wdk.getEntities({
+      ids: [wikidataid],
+      props: ['sitelinks'],
+      format: 'json',
+    })
+    const sitelinks = await fetch(entityAPIurl)
+    let sitelink = await sitelinks.json()
+    sitelink = sitelink.entities[wikidataid].sitelinks
+    if (sitelink && sitelink['enwiki']) {
```
jayvdb: 'enwiki' is choosing language (and repeated below); the site will soon be i18n's, so this needs to be in a variable which can be altered easily.

But why filter here?  This is scraping. More data is better than none.  Push all the extracted data through to the next layer, normalising and improving data. (e.g. `enwiki->` -> `wikipedia->en->` ?)

Then the UI templates choose which sitelinks they want to display.

jayvdb (dismissed):

lib/scrape.js
nalinbhardwaj: Oh okay, so the i18n is expected to work in frontend only? I thought it would be something like generating different packaged websites and choosing on display. But that works. I'll do all the changes

nalinbhardwaj:
@jayvdb : fixed
@blazeu : squashed

Issue #27, opened 2017-12-06
Reduce number of GitHub API hits


jayvdb:
In the initial GitHub org detection (78d0eaa), the following orgs were resolved using a GitHub search, as is recorded in the [build log](https://travis-ci.org/coala/gci-leaders/builds).

```
Could not find GitHub org for Apertium. Resorting to GitHub API hit.
Could not find GitHub org for BRL-CAD. Resorting to GitHub API hit.
Could not find GitHub org for Catrobat. Resorting to GitHub API hit.
Could not find GitHub org for CCExtractor Development. Resorting to GitHub API hit.
Could not find GitHub org for Drupal. Resorting to GitHub API hit.
Could not find GitHub org for Haiku. Resorting to GitHub API hit.
Could not find GitHub org for LibreHealth. Resorting to GitHub API hit.
Could not find GitHub org for MetaBrainz Foundation. Resorting to GitHub API hit.
Could not find GitHub org for OpenMRS. Resorting to GitHub API hit.
Could not find GitHub org for OpenWISP. Resorting to GitHub API hit.
Could not find GitHub org for OSGeo. Resorting to GitHub API hit.
Could not find GitHub org for Sugar Labs. Resorting to GitHub API hit.
Could not find GitHub org for Ubuntu. Resorting to GitHub API hit.
Could not find GitHub org for Wikimedia. Resorting to GitHub API hit.
Could not find GitHub org for XWiki. Resorting to GitHub API hit.
```

This task is for you to find alternative approaches to identity the GitHub org, in order to avoid using a GitHub API hit for these org.

Your build must be rebased and have at least one less API hit than master.

jayvdb:
[GitMate.io](https://gitmate.io) thinks the contributor prediction failed because there are not enough contributors. Three contributors must have authored comments on at least three issues each.

jayvdb:
Closing as new more precise issues should be created for any remaining performance improvements.

PR #81, opened 2017-12-12
scrape.js: Reduce number of API calls

andrewda:
Once we get search results from GitHub, be it an organization account
or a student account, we know it will almost certainly stay constant.
This will use the data from previous builds as a base to build off of
for future builds. Once we get data for a certain organization or
user, that data won't be searched for again.

Closes https://github.com/coala/gci-leaders/issues/27

### DIFF
--- a/lib/gather.js
+++ b/lib/gather.js
@@ -1,0 +1,13 @@
+const fetch = require('node-fetch')
+const fs = require('fs')
+
+async function getData() {
+  const res = await fetch(
+    'https://raw.githubusercontent.com/coala/gci-leaders/gh-pages/data.min.json'
+  )
+
+  const data = await res.json()
+  fs.writeFileSync(`${__dirname}/../out/data.min.json`, JSON.stringify(data))
+}
+
+getData()
--- a/lib/scrape.js
+++ b/lib/scrape.js
@@ -39,6 +39,15 @@ const GH_WEB_OPTIONS = {
   compress: false,
 }
 
+let existingData = []
+try {
+  existingData = JSON.parse(
+    fs.readFileSync(`${__dirname}/../out/data.min.json`).toString()
+  )
+} catch (e) {
+  console.log('No existing data...')
+}
+
 async function fetchProgram() {
   const res = await fetch(`${GCI_API_BASE}/program/2017/`)
   return await res.json()
@@ -209,12 +218,12 @@ async function findWiki(name) {
   }
 }
 
-async function findGitHubUser({ display_name }, org) {
+async function findGitHubUser(displayName, org) {
   if (!org) return
 
-  const shortName = validUsername(display_name)
+  const shortName = validUsername(displayName)
 
-  const username = await findGitHubUserInOrg(display_name, org)
+  const username = await findGitHubUserInOrg(displayName, org)
   if (username) return username
 
   const displayNamePattern = /^[a-zA-Z0-9-]{1,39}$/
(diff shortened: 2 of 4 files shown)


andrewda:
Looks like this is still WIP.

blazeu:
Pull request should start with `WIP:` if it's WIP.

.travis.yml
```diff
@@ -11,6 +11,7 @@ before_install:
 script:
   - coala --non-interactive -V
   - npm run build
+  - ./.ci/gather_data.sh
```
jayvdb: could this be JavaScript, so it can be incorporated into the npm build process ?

andrewda: I currently have it incorporated in the build script when `npm run gather` is run. Would it still be preferable to have it in JavaScript?

jayvdb: Yes, javascript, so it can be moved and re-used

package.json
```diff
@@ -4,7 +4,12 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "build": "node ./lib/scrape.js && node ./lib/generate.js"
+    "scrape": "node ./lib/scrape.js",
+    "generate": "node ./lib/generate.js",
+    "gather": "node ./lib/gather.js",
+    "clean": "rm -rf out && mkdir -p out && touch out/.keep",
```
blazeu: Remove the contents instead of the folder
i.e. `rm -rf out/*` 
and if you do that you don't need to recreate the `.keep` file

package.json
```diff
@@ -4,7 +4,12 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "build": "node ./lib/scrape.js && node ./lib/generate.js"
+    "scrape": "node ./lib/scrape.js",
+    "generate": "node ./lib/generate.js",
+    "gather": "node ./lib/gather.js",
+    "clean": "mkdir -p out && rm -rf out/*",
```
blazeu: I believe you don't need `mkdir -p out`  since the directory never got deleted.

lib/scrape.js
```diff
@@ -210,11 +220,14 @@ async function findWiki(name) {
 }
 
 async function findGitHubUser({ display_name }, org) {
+=======
+async function findGitHubUser(displayName, org) {
+>>>>>>> scrape.js: Reduce number of API calls
```
blazeu: ;)

andrewda: Shhh... 😋

blazeu:
ack https://github.com/coala/gci-leaders/pull/81/commits/79ca175bb1209bfb36e918155a20b5307d79db91

lib/gather.js
```diff
@@ -0,0 +1,23 @@
+const fetch = require('node-fetch')
+const fs = require('fs')
+
+function getRepository() {
+  if (process.env.REPOSITORY_URL) {
+    return process.env.REPOSITORY_URL.replace('git@github.com:', '')
```
jayvdb: Use the git remote ; then it works everywhere, Netlify , Travis, etc

https://www.npmjs.com/search?q=git+remote

blazeu:
ack https://github.com/coala/gci-leaders/pull/81/commits/79ca175bb1209bfb36e918155a20b5307d79db91

blazeu:
Replace `Closes` to `Related to`

This is a multiple instance task ;)

jayvdb (changes requested):
Also, if a forks `data.min.json` doesnt exist, it should fall back to using the main repos `data.min.json`.

Maybe even only the main repos `data.min.json` should be used, so that each patch can only be one "step" ahead of the main repo. (and then you can just hard code the main repos URL, no need to worry about finding the remote URL)

andrewda:
@jayvdb how do we determine the parent repo of the fork?

jayvdb:
> how do we determine the parent repo of the fork?

That would be another GitHub API call .. ;)

But I meant that it would be OK to just hard code `https://raw.githubusercontent.com/coala/gh-pages/data.min.json` .

I dont feel particularly inclined to help someone create an alternative version without sending their improvements to this repo.  I am more concerned about the ability for development forks of this repo to build correctly, so newcomers can easily participate in this repo.

The problem is that if you use `https://raw.githubusercontent.com/andrewda/gh-pages/data.min.json` , and dont use `https://raw.githubusercontent.com/coala/gh-pages/data.min.json` , the build on the fork will be 'reset' , and will need to have several builds before it catches up to the data which has been found in `https://raw.githubusercontent.com/coala/gh-pages/data.min.json` .

So if you do use `https://raw.githubusercontent.com/xxxx/gh-pages/data.min.json` , you should also fall back to `https://raw.githubusercontent.com/coala/gh-pages/data.min.json` (and IMO , just hard-code it).

jayvdb (approved):

Issue #38, opened 2017-12-06
Add i18n


jayvdb:
Include at least a few i18n strings, but this task is not about doing translations.

jayvdb:
I think we will want translations to be done on https://hosted.weblate.org/ , but we can set that up after basic functionality is merged.

mooncos:
Ok but can we use a hash table for the strings as of now?

jayvdb:
> Ok but can we use a hash table for the strings as of now?

Yes, we'll incrementally improve.  I am pretty sure weblate knows how to populate jquery.i18n data files

PR #78, opened 2017-12-12
Get better Lighthouse score in a11y and performance

myfrom:
I did a few things (all mentioned in the commits) to improve the Lighthouse score. There are a few differences between my tests and the test file from https://gci-leaders.netlify.com/ that were caused by my development environment and the fact that machine is fairly slow.

I also changed the output of /static from /out/static to the root of /out as we were discussing in #55

Here are test results before and after:
[tests.zip](https://github.com/coala/gci-leaders/files/1552817/tests.zip)

Related to #55, part of [GCI task](https://codein.withgoogle.com/dashboard/task-instances/5311708543320064/)

Issue #6, opened 2017-12-05
Add LICENSE


blazeu:
This repo uses [MIT license](https://github.com/coala/gci-leaders/blob/master/package.json#L11)

vigzmv:
I want to take this. Please assign it to me.

ayan-b:
@blazeu @jayvdb Is it still a GCI task? In case it is not I would like to add a LICENSE.md file.

blazeu:
@ayan-b You can take it.

@vigzmv unassigned because of inactivity.

blazeu:
@ayan-b Please join the org first.

ayan-b:
Sure. I will let you know once I join the org.

UPDATE: I have joined the org.

ayan-b:
Hi @blazeu, as it turns out, the maintainers are quite busy and hence I am unable to join the org. Can I still contribute without joining?

jayvdb:
@gitmate-bot assign

PR #84, opened 2017-12-13
i18n: Add full i18n support

mooncos:
This adds i18n support for the app, being
rendered in client-side so that translations
can be loaded dynamically. Implemented language
fallbacks for non-existing translations, plurals
and dynamic placeholders for strings.

Fixes https://github.com/coala/gci-leaders/issues/38

### DIFF
--- a/.coafile
+++ b/.coafile
@@ -9,7 +9,7 @@ bears = SpaceConsistencyBear
 default_actions = *: ApplyPatchAction
 
 [all.linelength]
-ignore += **.html
+ignore += **.html, static/js/i18n/**.json
 bears = LineLengthBear
 
 [all.links]
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -114,3 +114,7 @@ footer {
   color: grey;
   display: inline-block;
 }
+
+.chooser {
+  margin-top: 0.5em;
+}
--- a/static/js/global.js
+++ b/static/js/global.js
@@ -1,0 +1,47 @@
+var browserLocale
+
+if (!localStorage.getItem('lang')) {
+  var nav = window.navigator
+  browserLocale =
+    nav.language.split('-')[0] || nav.language || nav.userLanguage || ''
+} else {
+  browserLocale = localStorage.getItem('lang')
+}
+
+function updateTranslation(localex) {
+  $.i18n({
+    locale: localex,
+  })
+
+  $.i18n()
+    .load('./js/i18n', localex)
+    .done(function() {
+      console.log('i18n:' + localex + ' locale loaded')
+      $('body').i18n()
+      $('html').attr('lang', localex)
+      $.getScript('js/app.js')
+    })
+}
+
+$(window).on('load', function() {
+  var localeOptions = {
+    English: 'en',
+    Español: 'es',
+  }
+
+  var locList = $('#lang-select')
+  $.each(localeOptions, function(key, value) {
+    locList.append(
+      $('<option></option>')
+        .attr('value', value)
+        .text(key)
+    )
+  })
+
+  updateTranslation(browserLocale)
+  locList.val(browserLocale)
+  locList.on('change', function() {
+    updateTranslation(this.value)
+    localStorage.setItem('lang', this.value)
+  })
+})
--- a/templates/main.html
+++ b/templates/main.html
@@ -8,15 +8,26 @@
     <meta property="og:description" content="Google Code-in 2017 Current Leaders">
     <meta property="og:image" content="{{{rootURL}}}/images/GCI-logo.jpg">
     <link rel="stylesheet" href="css/main.css">
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
+    <!-- i18n libraries -->
+    <script src="https://cdn.rawgit.com/santhoshtr/CLDRPluralRuleParser/8baf9aedc4/src/CLDRPluralRuleParser.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery.i18n/1.0.4/jquery.i18n.min.js"></script>
+    <script src="https://cdn.rawgit.com/wikimedia/jquery.i18n/c932076e/src/jquery.i18n.messagestore.js"></script>
+    <script src="https://cdn.rawgit.com/wikimedia/jquery.i18n/c932076e/src/jquery.i18n.fallbacks.js"></script>
+    <script src="https://cdn.rawgit.com/wikimedia/jquery.i18n/c932076e/src/jquery.i18n.language.js"></script>
+    <script src="https://cdn.rawgit.com/wikimedia/jquery.i18n/c932076e/src/jquery.i18n.parser.js"></script>
+    <script src="https://cdn.rawgit.com/wikimedia/jquery.i18n/c932076e/src/jquery.i18n.emitter.js"></script>
+    <!-- global translation engine -->
+    <script src="js/global.js"></script>
   </head>
   <body>
     <div>
       <a href="http://codein.withgoogle.com">
         <img class="gci-logo" src="images/logos/gci.png" alt="Code-In logo">
       </a>
-      <h1>Google Code-in 2017 Current Leaders</h1>
+      <h1 data-i18n="page-title">Google Code-in 2017 Current Leaders</h1>
     </div>
-    <small id="time" data-time="{{datetime}}">Last updated: {{datetime}} <span id="ago"></span></small>
+    <small id="time" data-time="{{datetime}}"><span data-i18n="last-updated">Last updated</span>: {{datetime}} <span id="ago"></span></small>
     <div>
       <div class="progress-outer">
         <div class="progress-inner">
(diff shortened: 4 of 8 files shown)


.coafile
```diff
@@ -1,6 +1,6 @@
 [all]
 files = **.js, **.json, **.css, **.yml, **.html, .eslintrc
-ignore = node_modules/**, out/**, package-lock.json
+ignore = node_modules/**, out/**, package-lock.json, static/js/lib/**
```
blazeu: Don't do this please, we want those files to be linted

blazeu (changes requested):
Rather than storing the vendor js files in our repo, load it from somewhere, like cdnjs, jsdelivr or unpkg.

This will avoid the need to ignore those files in our linter, and it's the best that we can do before implementing a build system that takes care of npm/js package in the browser.

static/js/i18n/hi.json
```diff
@@ -0,0 +1,21 @@
+{
+  "page-title": "Google Code-in 2017 वर्तमान नेता",
+  "last-updated": "आखरी अपडेट",
+  "order-info": "प्रत्येक संगठन (और उनके GitHub खाते, यदि लागू हो) के लिए प्रमुख प्रतिभागी क्रमशः सूचीबद्ध होते हैं",
+  "tasks-completed": "पूरा कार्य",
+    "footer-text": "Google Code-in और इसके लोगो Google Inc. के ट्रेडमार्क हैं",
```
jayvdb: indent problem.

static/js/i18n/id.json
```diff
@@ -0,0 +1,21 @@
+{
+  "page-title": "Google Code-in 2017 Pemimpin Saat Ini",
+  "last-updated": "Terakhir diperbarui",
+  "order-info": "Peserta utama untuk setiap organisasi (dan akun GitHub mereka, jika ada) terdaftar secara acak.",
+  "tasks-completed": "Tugas selesai",
+  "footer-text": "Google Code-in dan logonya adalah merek dagang terdaftar dari Google Inc.",
+  "orgs-no-listed": "Organisasi tanpa pemimpin yang terdaftar",
+  "passed": "$1% passed, approximately $2 before task claiming ends",
```
jayvdb: if it isnt translated, it should be omitted, so that we can build checks for percentage translated, and tools to create translations.

does `jquery.i18n` not support fallback to English for missing messages?

If it doesnt, please dont include languages which you dont know your translation is 100% accurate and complete.  We can use http://hosted.weblate.org/ to manage the collection of translations and ensure they are complete before the file is available for the i18n layer.

static/js/i18n/hi.json
```diff
@@ -0,0 +1,21 @@
+{
+  "page-title": "Google Code-in 2017 वर्तमान नेता",
+  "last-updated": "आखरी अपडेट",
```
jayvdb: where did you get this translation from?

Is that translation MIT licensed per https://github.com/coala/gci-leaders/issues/6?
If not, please remove, and we'll acquire the translation again using a method that complies with the intended license of this repo.

jayvdb (changes requested):

static/js/i18n/hi.json
blazeu: IMO we should remove all translation for now and only keep `en` or 2.

.coafile
```diff
@@ -9,7 +9,7 @@ bears = SpaceConsistencyBear
 default_actions = *: ApplyPatchAction
 
 [all.linelength]
-ignore += **.html
+ignore += **.html, static/js/i18n/**.json
```
jayvdb, line 12: ugh?  Why is this?

They should all be unicode, and should all be nicely structured JSON.

templates/main.html
```diff
@@ -45,49 +56,48 @@ <h3>
...
                   src="https://assets-cdn.github.com/images/modules/logos_page/GitHub-Mark.png"
                   class="github"
-                  alt="Find them on GitHub"
                 />
               </a>
             {{/github}}
             {{#mailing_list}}
               <a href="{{mailing_list}}">
-                <img src="images/mail.png" class="chat" alt="Join their mailing list" />
+                <img src="images/mail.png" class="chat" />
```
jayvdb: why removing all the `alt`?
`jquery.i18n` doesnt support them?
English only `alt` is better than no `alt`..?

templates/main.html
```diff
@@ -1,22 +1,33 @@
 <!DOCTYPE html>
-<html lang="en">
+<html>
   <head>
     <meta charset="utf-8">
     <title>GCI Leaders</title>
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
```
jayvdb: This looks like it is doing something unrelated to the issue you are solving..?
please correct me if I am wrong about that.

templates/main.html
```diff
@@ -1,22 +1,33 @@
 <!DOCTYPE html>
-<html lang="en">
+<html>
```
jayvdb: lang="mul".

If not, it should still be `lang=en`, because it is only with JavaScript enabled that it is multilingual, and the JavaScript should update the lang attribute as required.

jayvdb (dismissed):

templates/main.html
jayvdb: Looks like it was related to https://github.com/coala/gci-leaders/pull/78/files#diff-c8935f416a66e3b982a42dccd8206926L6

templates/main.html
jayvdb: may also be related to the #78 merge

jayvdb:
unack https://github.com/coala/gci-leaders/pull/84/commits/911e838cd0a3d2c561dd6a50b6815e58b83d3953

mooncos:
Sorry those were conflicts with rebasing from master as `main.html` was changed. Will fix now.

.coafile
mooncos: JSON doesn't have linebreaks and the translations for some strings are too long for the linelength limit.

templates/main.html
```diff
@@ -182,9 +193,13 @@ <h3>
       </a>
     </div>
     <footer>
-      <small>Google Code-in and the Google Code-in logo are trademarks of Google Inc.</small>
+      <small data-i18n="footer-text">Google Code-in and the Google Code-in logo are trademarks of Google Inc.</small>
     </footer>
-    <script src="js/app.js"></script>
+    <script>
+      $(window).on('load', function() {
+        $.getScript('js/app.js')
```
blazeu: Whaat ..?

blazeu: Did the previous `defer` not work?

blazeu:
Can you add a languange switcher? Maybe using `<select>`

static/js/global.js
```diff
@@ -0,0 +1,24 @@
+var browser_locale = window.navigator.language.split('-')[0]
```
blazeu: camelCase please

static/js/global.js
```diff
@@ -0,0 +1,24 @@
...
+      $('html').attr('lang', localex)
+      $.getScript('js/app.js')
+    })
+}
+
+$(window).on('load', function() {
+  updateTranslation(browser_locale)
+  $('#lang-select').val(browser_locale)
+  $('#lang-select').on('change', function() {
+    updateTranslation(this.value)
```
blazeu: It would be nice if the user's preference is saved (preferably in localStorage), so that the user doesn't need to change it on every visit.

blazeu (changes requested):

static/js/global.js
blazeu: Also why is this not in the other file? (language_switcher)

static/js/global.js
mooncos: oops too accustomed to python hehe

templates/main.html
```diff
@@ -182,9 +194,14 @@ <h3>
...
```
blazeu: I saw you were using `defer` before, why change?

mooncos: `defer` was not working properly. I implemented this workaround to make sure that the translation file finished importing before applying the translations

static/css/main.css
```diff
@@ -114,3 +114,8 @@ footer {
   color: grey;
   display: inline-block;
 }
+
+.chooser {
+  margin-top: 0.5em;
+  display: flex;
```
blazeu: I don't think you need `flex`

templates/main.html
blazeu: >translation file finished importing

Then it's not `$(window).on('load')`, there must be another event from jquery.i18n

static/js/global.js
```diff
@@ -0,0 +1,24 @@
...
+    locale: localex,
+  })
+
+  $.i18n()
+    .load('./js/i18n', localex)
+    .done(function() {
+      console.log('i18n:' + localex + ' locale loaded')
+      $('body').i18n()
+      $('html').attr('lang', localex)
+      $.getScript('js/app.js')
```
blazeu, line 22: ^ You have done it here.

templates/main.html
mooncos: It needs to because the text to translate (the HTML) needs to load as well so `$(window).on('load')` is the correct way.

static/js/global.js
```diff
@@ -0,0 +1,24 @@
...
+    })
+}
+
+$(window).on('load', function() {
```
blazeu, line 26: `$(window).on('load')` is here ^

static/js/global.js
```diff
@@ -0,0 +1,48 @@
+var browserLocale
+
+if (!localStorage.getItem('lang')) {
+  var nav = window.navigator
+  browserLocale =
+    nav.language.split('-')[0] || nav.language || nav.userLanguage || ''
+  localStorage.setItem('lang', browserLocale)
```
blazeu: Only store if the user has specifically selected a language.

mooncos: ok

blazeu (approved):

blazeu:
ack https://github.com/coala/gci-leaders/pull/84/commits/2899de985a6aee5fb9c8c2d1c9f86fb796c5d678

Issue #65, opened 2017-12-11
Create blog planet from blog_url


jayvdb:
The data feed has a `blog_url` for many of the orgs:

- coala
- drupal
- FOSSASIA
- HaikuOS
- JBoss
- Liquid Galaxylab
- mifos
- musicbrainz
- OpenMRS
- osgeo
- systers
- ubuntu
- Wikimedia
- XWiki
- Zulip

Exclude URLs which do not have a properly declared RSS/ATOM alternative, e.g. BRL-CAD's FB url (https://fb.me/brlcad) and http://forum.terasology.org/ and http://www.scorelab.org/blog/index.html

jayvdb:
I suggest this page be an overlay / 'tab' , and maybe move the Twitter feed to also be an overlay / tab

PR #86, opened 2017-12-15
Create blog planet

9r0x:
This commit creates a blog planet from gci posts.

Closes https://github.com/coala/gci-leaders/issues/65

### DIFF
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -3,6 +3,7 @@ const Mustache = require('mustache')
 const ncp = require('ncp').ncp
 const orgs = require('../out/data.json')
 const dates = require('../out/dates.json')
+const planet = require('../out/blog_planet.json')
 
 const time = fs.statSync(`${__dirname}/../out/data.json`).mtime
 const datetime = new Date(time).toUTCString()
@@ -32,5 +33,14 @@ fs.writeFileSync(
     noLeader,
     competitionOpen,
     noClaims,
+  })
+)
+
+fs.mkdirSync(`${__dirname}/../out/planet`)
+
+fs.writeFileSync(
+  `${__dirname}/../out/planet/index.html`,
+  Mustache.render(fs.readFileSync('templates/planet.html').toString(), {
+    planet,
   })
 )
--- a/lib/planet.js
+++ b/lib/planet.js
@@ -1,0 +1,62 @@
+const data = require('../out/data.json')
+const feedFinder = require('find-rss')
+const fs = require('fs')
+const feedParser = require('feed-read-parser')
+const gciStart = new Date('November 27, 2017')
+
+const errs = []
+let articles = []
+
+function getFeeds(feedUrls, i, cb) {
+  let feedUrl = feedUrls[i]
+  if (!feedUrl) return cb(errs, articles)
+  feedParser.get(feedUrl, function(err, _articles) {
+    if (err) errs.push(err.message + ' ' + feedUrl)
+    else {
+      articles = articles.concat(_articles)
+    }
+    getFeeds(feedUrls, i + 1, cb)
+  })
+}
+
+async function getFeedLinks(data) {
+  const linkPromises = data.map(org => {
+    const { blog_url } = org
+    if (blog_url) {
+      return feedFinder(blog_url)
+        .catch(err => console.error(err))
+        .then(res => {
+          if (res.length > 0) {
+            return res[0].url
+          }
+        })
+    }
+  })
+  return await Promise.all(linkPromises)
+}
+
+function wordInString(s, word) {
+  return new RegExp('\\b' + word + '\\b', 'i').test(s)
+}
+
+;(async () => {
+  const feedLinks = (await getFeedLinks(data)).filter(link => link)
+  getFeeds(feedLinks, 0, (errs, articles) => {
+    if (errs.length > 0) {
+      errs.forEach(err => console.error(err))
+    }
+    articles.forEach(
+      article => (article.published = new Date(article.published))
+    )
+    articles = articles.filter(
+      article =>
+        article.published > gciStart ||
+        wordInString(article.content, 'Google Code-In')
+    )
+    articles.sort((a, b) => b.published - a.published)
+    fs.writeFileSync(
+      `${__dirname}/../out/blog_planet.json`,
+      JSON.stringify(articles)
+    )
+  })
+})()
--- a/.coafile
+++ b/.coafile
@@ -9,7 +9,7 @@ bears = SpaceConsistencyBear
 default_actions = *: ApplyPatchAction
 
 [all.linelength]
-ignore += **.html, static/js/i18n/**.json
+ignore += **.html, static/js/i18n/**.json, package.json
 bears = LineLengthBear
 
 [all.links]
--- a/package.json
+++ b/package.json
@@ -4,18 +4,21 @@
   "description": "",
   "main": "index.js",
   "scripts": {
+    "planet": "node ./lib/planet.js",
     "scrape": "node ./lib/scrape.js",
     "generate": "node ./lib/generate.js",
     "gather": "node ./lib/gather.js",
     "clean": "rm -rf out/*",
-    "build": "npm run gather && npm run scrape && npm run generate",
-    "build:clean": "npm run clean && npm run scrape && npm run generate"
+    "build": "npm run gather && npm run scrape && npm run planet && npm run generate",
+    "build:clean": "npm run clean && npm run build"
   },
   "keywords": [],
   "author": "blazeu <neversleepman@tfwno.gf>",
   "license": "MIT",
   "dependencies": {
     "chattie": "0.0.4",
+    "feed-read-parser": "^0.0.6",
+    "find-rss": "^1.6.4",
     "json2yaml": "^1.1.0",
     "mustache": "^2.3.0",
     "ncp": "^2.0.0",
--- a/templates/planet.html
+++ b/templates/planet.html
@@ -1,0 +1,27 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Google Code-in Blog Planet</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+  </head>
+  <body>
+    <div>
+      <h1>Welcome to the Google Code-In Blog Planet</h1>
+      <hr />
+    </div>
+    <div>
+      {{#planet}}
+      <a class="title" href={{link}}><h1>{{title}}</h1></a>
+      <h5>{{published}}</h5>
+      <i class="author">{{author}}</i>
+      <div class="content">
+        {{{content}}}
+      </div>
+      {{/planet}}
+    </div>
+    <footer>
+      <small>Google Code-in and the Google Code-in logo are trademarks of Google Inc.</small>
+    </footer>
+  </body>
+</html>


lib/scrape.js
```diff
@@ -48,6 +49,23 @@ try {
...
+  'https://blog.zulip.org/rss/',
+  'https://blog.musicbrainz.org/feed/',
+  'https://insights.ubuntu.com/feed/',
+  'http://www.xwiki.org/xwiki/bin/view/Main/WebRss?xpage=rdf',
+  'https://www.drupal.org/planet/rss.xml',
+  'http://planet.osgeo.org/atom.xml',
+  // Not formateed XML 'https://blog.coala.io/atom.xml',
+  'http://mifos.org/feed/',
+  'https://blog.fossasia.org/feed/',
+  'http://openmrs.org/feed/',
```
blazeu: This has to come from `blog_url` not hardcoded.

blazeu (changes requested):

lib/scrape.js
9r0x: The reason I hardcoded is that coala's feed link has some problem(it is not formatted well) and it breaks the program.

lib/scrape.js
```diff
@@ -48,6 +49,23 @@ try {
...
+]
```
andrewda: Ideally these would be fetched automatically, right? Using the `blog_url` key from the API and scraping for something like

```html
<link rel="alternate" type="application/atom+xml" href="/atom.xml" title="Feed">
```
edit: (nevermind, @blazeu beat me to it 😋 )

9r0x: Right, I did fetch them, but there is a problem with coala's feed, can you check `https://blog.coala.io/atom.xml` on your machine?

lib/scrape.js
blazeu: Then make the program not break :P

9r0x: OK:)
Can I ignore coala's feed for now?  I opened the xml on my machine, it is empty.

lib/scrape.js
andrewda: Yep, seems to not be working. But in that case just skip over it and only use working XML feeds. That'll be encouragement to get the coala feed fixed 😉

lib/scrape.js
blazeu: Make it so that when the parser failed it's skipping instead of "breaking".

lib/scrape.js
jayvdb: I've created https://github.com/coala/devops/issues/92 to fix this, but that will be much later after this is merged.

9r0x: Seems it's fixed now😉

blazeu:
https://deploy-preview-86--gci-leaders.netlify.com/blog is empty or am I seeing it wrong here.

lib/generate.js
```diff
@@ -34,3 +35,24 @@ fs.writeFileSync(
...
+try {
+  fs.writeFileSync(
+    `${__dirname}/../out/blog/index.html`,
+    Mustache.render(fs.readFileSync('templates/blog.html').toString(), {
+      feeds,
+    })
+  )
+} catch (err) {
+  if (err.code === 'ENOENT') {
+    fs.mkdir(`${__dirname}/../out/blog`)
```
blazeu: This is crappy.

You can modify the build script to clean the folder everytime and then recreate the folder here everytime.

Also this is async and it's possible that the script below will run first. Use sync version of this.

9r0x:
Right, also https://deploy-preview-86--gci-leaders.netlify.com/feeds.json is empty

lib/scrape.js
```diff
@@ -331,6 +333,28 @@ async function fetchDates() {
   return res.json()
 }
 
+function getFeeds(feed_url, callback) {
+  if (feed_url instanceof Array) {
+    var feed_urls = feed_url,
+      articles = [],
+      errs = []
```
blazeu: Can you not copy and paste this :P (the function)

lib/scrape.js
```diff
@@ -340,6 +364,32 @@ async function fetchDates() {
...
+      if (blog_url != null && blog_url != '') {
+        return feedFinder(blog_url)
+          .catch(err => console.error(err))
+          .then(res => {
+            if (res.length > 0) {
+              res[0].url
+            }
+          })
+      }
+    })
```
blazeu: Probably looks neater if you put it inside variable first and then use `Promise.all()`.

Also, async function please, instead of `then`.

lib/scrape.js
```diff
@@ -340,6 +364,32 @@ async function fetchDates() {
     (a, b) => b.completed_task_instance_count - a.completed_task_instance_count
   )
 
+  Promise.all(
+    data.map(org => {
+      const { blog_url } = org
+      if (blog_url != null && blog_url != '') {
```
blazeu: `if (blog_url)` is enough, I think.

blazeu (changes requested):

lib/scrape.js
```diff
@@ -340,6 +364,32 @@ async function fetchDates() {
     (a, b) => b.completed_task_instance_count - a.completed_task_instance_count
   )
 
+  Promise.all(
```
blazeu: Probably looks neater if you put it inside variable first and then use `Promise.all()`.

Also, async function please, instead of `then`.

blazeu (changes requested):

blazeu:
Maybe put feed related stuff in it's own file.

The thing in generate.js can stay there, but scrape.js is getting very long.

lib/scrape.js
```diff
@@ -331,6 +333,28 @@ async function fetchDates() {
   return res.json()
 }
 
+function getFeeds(feed_url, callback) {
```
jayvdb: `feed_url` -> `feed_urls`

jayvdb (changes requested):
blog entries need to be sorted by date published in reverse order, and could you also eliminate blog posts before Nov 27.  Possibly also include earlier blog posts, *if* they mention Google Code-in.  main reason is the page is too big to read, with all those lovely GSOC blog posts from students.

lib/feed.js
```diff
@@ -0,0 +1,45 @@
+const data = require('../out/data.json')
+const feedFinder = require('find-rss')
+const fs = require('fs')
+const feedParser = require('feed-read-parser')
+
+const errs = []
+var articles = []
+
+function getFeeds(feed_urls, i, cb) {
+  var feed_url = feed_urls[i]
```
blazeu: `const` / `let` only please (for consistency).

blazeu: Also camelCase

lib/feed.js
```diff
@@ -0,0 +1,45 @@
...
+            return res[0].url
+          }
+        })
+    }
+  })
+  return await Promise.all(linkPromises)
+}
+
+;(async () => {
+  const feedLinks = (await getFeedLinks(data)).filter(link => link != undefined)
```
blazeu: if you're checking for `null` **and** `undefined` **and** empty string, just `link => link` is enough.

lib/feed.js
```diff
@@ -0,0 +1,45 @@
...
+  getFeeds(feedLinks, 0, function(errs, articles) {
+    if (errs.length > 0) {
+      errs.map(err => console.error(err))
```
blazeu: `map` returns new array, use `forEach`

lib/feed.js
```diff
@@ -0,0 +1,45 @@
...
+          }
+        })
+    }
+  })
+  return await Promise.all(linkPromises)
+}
+
+;(async () => {
+  const feedLinks = (await getFeedLinks(data)).filter(link => link != undefined)
+  getFeeds(feedLinks, 0, function(errs, articles) {
```
blazeu: You might wanna use arrow function

lib/generate.js
```diff
@@ -34,3 +36,12 @@ fs.writeFileSync(
     noClaims,
   })
 )
+
+mkdirp(`${__dirname}/../out/blog`)
```
blazeu: Again, this is async and the writeFile below might run first before this is finished.
There is a sync version of this.

blazeu: But more importantly, why not `fs.mkdirSync` ?

lib/feed.js
```diff
@@ -0,0 +1,55 @@
...
+  getFeeds(feedLinks, 0, (errs, articles) => {
+    if (errs.length > 0) {
+      errs.foreach(err => console.error(err))
+    }
+    const feeds = articles.filter(
+      feed =>
+        (new Date(feed.published) > new Date('November 27, 2017')) |
+        wordInString(feed.content, 'Google Code-In')
+    )
+    feeds.sort((a, b) => new Date(b.published) - new Date(a.published))
```
jayvdb: this creates lots of Date objects (and also created above to be compared with `Date('November 27, 2017')`).

As all `published` will need to be compared, create Date objects for them all once, before sorting.

lib/feed.js
```diff
@@ -0,0 +1,55 @@
...
+
+;(async () => {
+  const feedLinks = (await getFeedLinks(data)).filter(link => link)
+  getFeeds(feedLinks, 0, (errs, articles) => {
+    if (errs.length > 0) {
+      errs.foreach(err => console.error(err))
+    }
+    const feeds = articles.filter(
+      feed =>
+        (new Date(feed.published) > new Date('November 27, 2017')) |
```
jayvdb: GCI start date is a constant that should be moved to the top of the file.

lib/feed.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+;(async () => {
+  const feedLinks = (await getFeedLinks(data)).filter(link => link)
+  getFeeds(feedLinks, 0, (errs, articles) => {
+    if (errs.length > 0) {
+      errs.foreach(err => console.error(err))
+    }
+    const feeds = articles.filter(
+      feed =>
```
jayvdb: the name `feed` here is confusing.  A feed is a list of articles, so "`feed.published`" hurts my brain.

lib/feed.js
```diff
@@ -0,0 +1,55 @@
...
+        (new Date(feed.published) > new Date('November 27, 2017')) |
+        wordInString(feed.content, 'Google Code-In')
+    )
+    feeds.sort((a, b) => new Date(b.published) - new Date(a.published))
+    fs.writeFileSync(`${__dirname}/../out/feeds.json`, JSON.stringify(feeds))
```
jayvdb: this is creating *one* new feed, not "feeds".  And it contains a specific type of articles, so lets call it "blog_planet.json".

After this PR is merged, please create a new issue to emit the feed as an RSS or ATOM file, so it can be read using a RSS reader.  And that file should be called  `blog_planet.rss` or `blog_planet.atom`.

templates/blog.html
```diff
@@ -0,0 +1,26 @@
...
+  </head>
+  <body>
+    <div>
+      <h1>Welcome to the Google Code-In Blog Planet</h1>
+      <hr />
+    </div>
+    <div>
+      {{#feeds}}
+      <a class="title" href={{link}}><h3>{{title}}</h3></a>
+      <i class="author">{{author}}</i>
```
jayvdb: add date published for each entry

jayvdb (changes requested):

lib/feed.js
jayvdb: Also rename `lib/feed.js` to `lib/planet.js` .

"gci-leaders" may also start publishing its own RSS feed (https://github.com/coala/gci-leaders/issues/91)

9r0x:
https://deploy-preview-86--gci-leaders.netlify.com/planet

lib/planet.js
```diff
@@ -0,0 +1,60 @@
+const data = require('../out/data.json')
+const feedFinder = require('find-rss')
+const fs = require('fs')
+const feedParser = require('feed-read-parser')
+const gciStart = new Date('November 27, 2017')
+
+const errs = []
+let articles = []
```
blazeu, line 8: This isn't used outside of the function right?

lib/planet.js
```diff
@@ -0,0 +1,60 @@
...
+  return new RegExp('\\b' + word + '\\b', 'i').test(s)
+}
+
+;(async () => {
+  const feedLinks = (await getFeedLinks(data)).filter(link => link)
+  getFeeds(feedLinks, 0, (errs, articles) => {
+    if (errs.length > 0) {
+      errs.foreach(err => console.error(err))
+    }
+    articles.map(article => (article.published = new Date(article.published)))
```
blazeu: Did you mean `forEach`? `map` is for creating a new array.

blazeu: the proper way of using `map` is

`articles = articles.map(..`

lib/planet.js
blazeu: There's a better way of doing recursion than this. But it's ok for now, no changes.

You don't need recursion even, a loop is better.

9r0x:
@blazeu Actually I write that `map` intentionally, though I don't know why it behaves this way, see the following screenshots pls.

If there are better ways to achieve this, please tell me:P

screen shot 2017-12-18 at 2 13 42 pm
screen shot 2017-12-18 at 2 13 58 pm

blazeu:
It's because you're mutating the object directly (`article.`)
You'll get the same effect with `forEach` without creating a new array that is not used.

blazeu:
If you want to use `var = arr.map` then you need to return the whole object, not just mutating them.

But let's just use `forEach`

lib/planet.js
```diff
@@ -0,0 +1,62 @@
...
+  getFeeds(feedLinks, 0, (errs, articles) => {
+    if (errs.length > 0) {
+      errs.forEach(err => console.error(err))
+    }
+    articles.forEach(
+      article => (article.published = new Date(article.published))
+    )
+    articles = articles.filter(
+      article =>
+        (article.published > gciStart) |
```
blazeu: `||`

`|` is bitwise or

jayvdb (approved):

blazeu (approved):

blazeu:
ack https://github.com/coala/gci-leaders/pull/86/commits/ae85e4ae70d021f984205335bc148b789afec522

Issue #93, opened 2017-12-18
Handle username change / deletion


blazeu:
Leader of MetaBrainz Foundation, Eshan Singh has changed their GitHub username.

As a result, the GitHub link in gci-leaders is broken.

PR #94, opened 2017-12-18
scrape.js: Handle username changes and deletions

andrewda:
Prior to this, if a user's GitHub account was cached, it would be
assumed to be correct for all future builds. However, this is not
the case if the user has changed or deleted their account. This
will verify that the cached username exists before using it, and
rerun the username finding code if it does not exist.

Closes https://github.com/coala/gci-leaders/issues/93

### DIFF
--- a/lib/scrape.js
+++ b/lib/scrape.js
@@ -5,12 +5,16 @@ const json2yaml = require('json2yaml')
 const validUsername = require('valid-github-username')
 const wdk = require('wikidata-sdk')
 
-const GH_USER_BASE = 'https://github.com/users'
-const GH_ORG_BASE = 'https://github.com/orgs'
+const GH_BASE = 'https://github.com'
+const GH_USER_BASE = `${GH_BASE}/users`
+const GH_ORG_BASE = `${GH_BASE}/orgs`
 const GH_API_BASE = 'https://api.github.com'
 const GCI_API_BASE = 'https://codein.withgoogle.com/api'
 
 const MIN_SEARCH_SCORE = 10
+
+// The time to cache GitHub usernames for in milliseconds
+const GITHUB_CACHE_TIME = 7 * 24 * 60 * 60 * 1000
 
 const CHAT_IMAGES = {
   GITTER: 'images/logos/gitter.png',
@@ -63,6 +67,11 @@ async function fetchLeaders(id) {
   const res = await fetch(`${GCI_API_BASE}/program/current/organization/${id}`)
   const { leaders } = await res.json()
   return leaders
+}
+
+async function checkGitHubUserExists(user) {
+  const res = await fetch(`${GH_BASE}/${user}`)
+  return res.status === 200
 }
 
 async function searchGitHubOrgs(query) {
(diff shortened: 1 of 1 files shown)


lib/scrape.js
```diff
@@ -65,6 +66,11 @@ async function fetchLeaders(id) {
   return leaders
 }
 
+async function checkGitHubUserExists(user) {
+  const res = await fetch(`${GH_BASE}/${user}`)
+  return res.status === 200 ? true : false
```
blazeu: === will return boolean, no need to do ternary.

blazeu (changes requested):

lib/scrape.js
```diff
@@ -224,7 +233,12 @@ async function findGitHubUser(displayName, org) {
   const shortName = validUsername(displayName)
 
   const username = await findGitHubUserInOrg(displayName, org)
-  if (username) return username
+  if (username) {
+    return {
+      login: username,
+      updated: Date.now(),
```
jayvdb: you're abusing this function.  A find function will always have an `updated` of "now", so it is redundant and annoying to be returning that to the caller.

the storing of cache information doesnt belong in this function.

lib/scrape.js
```diff
@@ -281,13 +298,33 @@ async function fetchOrgsWithData() {
...
     const existingOrg = existingData.find(existing => existing.id === org.id)
-    const fetchingUsers = orgLeaders[index].map(user => {
+    const fetchingUsers = orgLeaders[index].map(async user => {
       if (existingOrg && existingOrg.leaders) {
         const existingUser = existingOrg.leaders.find(
           leader => leader.id === user.id
         )
-        if (existingUser && existingUser.github_account) {
-          return existingUser.github_account
+        if (
```
jayvdb: move this out to a new function so as to not over-complicate this function with GitHub stuff , and also a new function will be testable.

jayvdb (changes requested):

lib/scrape.js
```diff
@@ -224,7 +233,9 @@ async function findGitHubUser(displayName, org) {
   const shortName = validUsername(displayName)
 
   const username = await findGitHubUserInOrg(displayName, org)
-  if (username) return username
+  if (username) {
```
jayvdb: unnecessary change here

lib/scrape.js
```diff
@@ -260,6 +271,35 @@ async function findGitHubUserInOrg(user, org) {
   return match ? match[1] : null
 }
 
+function checkCacheExpired(user) {
+  if (user && user.github_updated) {
+    return Date.now() - user.github_updated > GITHUB_CACHE_TIME
+  }
+}
+
+async function fetchUserFromCache(user, existingUser, organization) {
```
jayvdb: I would name this something like `freshenUserGitHubCache`

lib/scrape.js
```diff
@@ -260,6 +271,35 @@ async function findGitHubUserInOrg(user, org) {
   return match ? match[1] : null
 }
 
+function checkCacheExpired(user) {
```
jayvdb: Also mention GitHub in this function name

lib/scrape.js
```diff
@@ -281,25 +321,28 @@ async function fetchOrgsWithData() {
...
       if (existingOrg && existingOrg.leaders) {
         const existingUser = existingOrg.leaders.find(
           leader => leader.id === user.id
         )
-        if (existingUser && existingUser.github_account) {
-          return existingUser.github_account
+
+        return await fetchUserFromCache(user, existingUser, orgGitHub[index])
+      } else {
+        return {
```
jayvdb: this branch could also use `fetchUserFromCache`, by with `existingUser` being undefined.

jayvdb (approved):

blazeu (approved):

blazeu:
Just need a rebase.

blazeu:
ack https://github.com/coala/gci-leaders/pull/94/commits/209b9f482d0d3e3a97c01f6c2ebea7d17c50492c

Issue #91, opened 2017-12-18
Create RSS feed of data changes


jayvdb:
Compare the new data with the previously deployed data to manage an RSS (or ATOM) feed of changes occurring to the data.

e.g.

- How many tasks completed
- Change of order of the organisation
- Change of who is on the leaderboard

jayvdb:
Ideally the solution to this uses the deployed RSS as its data source of prior events, so the solution can prepend only new events to the RSS feed.

PR #95, opened 2017-12-19
Create RSS feed of data changes

andrewda:
This adds an RSS feed containing all the changes in the data.

Closes https://github.com/coala/gci-leaders/issues/91


andrewda:
Here's the current generated feed: https://deploy-preview-95--gci-leaders.netlify.com/feed.xml

lib/scrape.js
```diff
@@ -369,14 +377,19 @@ async function fetchDates() {
...
 
   // sort data by completed_task_instance_count
-  data.sort(
+  orgs.sort(
     (a, b) => b.completed_task_instance_count - a.completed_task_instance_count
   )
 
+  const data = {
+    orgs: sortKeys({ orgs }, { deep: true }).orgs,
+    lastUpdated: new Date(),
```
blazeu: This isn't used later, right?

rss.js script is reading from feed_items.json

But this is bad in general, we want the json structure to be stable from time to time, so that we can easily process it later and do some analytics in the end.

jayvdb:
Would be good to have the RSS mentioned in the `index.html`

lib/scrape.js
andrewda: This is used [here](https://github.com/coala/gci-leaders/pull/95/files#diff-9676cbcb5938ab810f7fa82a528abc4eR46) to ensure that we're not repeating the same data changes in the RSS. What's the best alternative you can think of?

blazeu: Maybe create a new file that holds the last updated time.

We have many generated files (org data, blog planet, dates) and that'll give the last updated time to all of that data.

lib/rss.js
```diff
@@ -0,0 +1,192 @@
+const RSS = require('rss')
+const generateDiff = require('deep-diff').diff
```
blazeu, line 2: Could use destructuring here

blazeu: ^ Ah, but that'll make it look crappier

lib/rss.js
```diff
@@ -0,0 +1,192 @@
...
+)
+
+const oldOrgs = createOrgHash(oldData)
+const newOrgs = createOrgHash(newData)
+
+console.log(oldOrgs)
+
+if (oldOrgs && newOrgs) {
+  const diffs = generateDiff(oldOrgs, newOrgs) || []
+  console.log(diffs)
```
blazeu: Do we need this? or is this for debugging, we might want a fancier log message or just throw this away completely.

lib/rss.js
```diff
@@ -0,0 +1,192 @@
...
+    elem.leaders = createLeadersHash(elem.leaders)
+    hash[elem.slug] = elem
+    return hash
+  }, {})
+}
+
+let current = {}
+try {
+  current = JSON.parse(
+    fs.readFileSync(`${__dirname}/../out/feed_items.json`).toString()
```
blazeu: We might want to fetch this like `data.json` or the file will not be available on every build.

lib/rss.js
```diff
@@ -0,0 +1,192 @@
...
+        title = `Organization Rank Updated for ${organization.name}`
+      } else {
+        title = `Updated "${
+          kind === 'A' ? itemPath[itemPath.length - 2] : finalProperty
+        }" for ${organization.name}`
+      }
+
+      if (kind === 'A') {
+        return feedItems.push({
+          title,
```
blazeu: Might want to add description.

The value can be a readable form of `property:old` and `property:new`

lib/rss.js
```diff
@@ -0,0 +1,192 @@
...
+          date: newData.lastUpdated,
```
blazeu: Should be `pubDate`

https://validator.w3.org/feed/docs/rss2.html#hrelementsOfLtitemgt

lib/rss.js
```diff
@@ -0,0 +1,192 @@
...
+
+console.log(oldOrgs)
+
+if (oldOrgs && newOrgs) {
+  const diffs = generateDiff(oldOrgs, newOrgs) || []
+  console.log(diffs)
+
+  const feedItems = current.items || []
+
+  if (current.lastUpdated !== buildInfo.generated) {
```
jayvdb: how can they ever be equal!?

more importantly, you do not need to store the date generated.  The `out/` files have dates, and HTTP tells you the dates of those files.

lib/rss.js
```diff
@@ -0,0 +1,192 @@
...
+
+const buildInfo = JSON.parse(
+  fs.readFileSync(`${__dirname}/../out/scrape_info.json`).toString()
+)
+
+const oldData = JSON.parse(
+  fs.readFileSync(`${__dirname}/../out/data_old.json`).toString()
+)
+const newData = JSON.parse(
+  fs.readFileSync(`${__dirname}/../out/data.json`).toString()
```
jayvdb: Every new .js is loading this file again.  And creating a new script , and a longer `build` command.

If these .js are modularised, the data can be parsed once and passed around .

jayvdb (dismissed):

lib/rss.js
blazeu: >HTTP tells you the dates of those files.

It does, but not all host will serve the correct date, e.g. netlify.

blazeu: But yeah, it'll never be equal.
So no need to read the date.

andrewda: The value of `current.lastUpdated` is set to `buildInfo.generated` [here](https://github.com/coala/gci-leaders/pull/95/files#diff-9676cbcb5938ab810f7fa82a528abc4eR183) when the file is created, so they will be equal if the same `data.json` file is being run multiple times.

lib/rss.js
andrewda: Should I modularize in this PR? Seems like a very big change. After this gets merged in, I could definitely do that in a new PR though so as not to clutter this one with too many changes.

lib/rss.js
andrewda: The `rss` module does this automatically for me. See the output: https://deploy-preview-95--gci-leaders.netlify.com/feed.xml

lib/rss.js
andrewda: Alright done now in `gather.js` file.

lib/rss.js
andrewda: Done! 👍

lib/rss.js
```diff
@@ -0,0 +1,203 @@
...
+try {
+  current = JSON.parse(
+    fs.readFileSync(`${__dirname}/../out/feed_items.json`).toString()
+  )
+} catch (e) {
+  current = {}
+}
+
+const buildInfo = JSON.parse(
+  fs.readFileSync(`${__dirname}/../out/scrape_info.json`).toString()
```
blazeu: So we don't need to fetch this from the network?

You can use the modified date from the json file, I think we do that for the "Last updated" information.

andrewda: Ok, so instead of having a `scrape_info.json` file we should check the modified time on the `data.json` file using `fs.stat()`?

blazeu: Yes, for now.

We'll modularize it once this is done.

lib/rss.js
jayvdb: yes, you should modularise the new code in this PR.

jayvdb: And we have tests now.  The bar has been raised :P

lib/rss.js
jayvdb: netlify should serve the correct date.  We need to set the date on the files so they stay constant even after regeneration , if the content is the same.

tests/lib/rss.test.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * RSS Library Test
+ */
+
+const fs = require('fs')
+
+describe('lib.rss', () => {
+  it('should create a feed.xml file', () => {
+    const path = `${__dirname}/../../out/feed.xml`
+    const file = fs.existsSync(path)
```
blazeu: Call the rss function itself with mock data.

blazeu (changes requested):

lib/rss.js
```diff
@@ -0,0 +1,192 @@
...
+      const feedData = {
+        items: feedItems,
+        lastUpdated: dataUpdated,
+      }
+
+      fs.writeFileSync(
+        `${__dirname}/../out/feed_items.json`,
+        JSON.stringify(feedData)
+      )
+      fs.writeFileSync(`${__dirname}/../out/feed.xml`, feed.xml())
```
blazeu: if this belongs in `main`, the test can be isolated without mocking fs calls

lib/rss.js
```diff
@@ -0,0 +1,192 @@
...
+    return hash
+  }, {})
+}
+
+module.exports = async ({
+  data: newData,
+  data_old: oldData,
+  feed_items: current,
+}) => {
+  const dataUpdated = fs.statSync(`${__dirname}/../out/data.json`).mtime
```
blazeu: This can come from the arg

tests/lib/rss.test.js
blazeu: `main` can export the function, beside running them directly, so that it can be called with mocked fs and then tested with this test.

lib/rss.js
andrewda: I didn't want to bloat the main script too much but sure I can do the stat from the main script.

tests/lib/rss.test.js
andrewda: Sure, I'll call the RSS function with mock data, but I don't think using `main` in the `rss` test is very good - defeats the purpose of unit tests. I'll make a main test when I finish refactoring in a new PR.

lib/rss.js
jayvdb: I agree; separating generation from the layout of the files is a good idea.

andrewda:
Changes addressed 😄

tests/lib/rss.test.js
```diff
@@ -0,0 +1,59 @@
...
+        fs.readFileSync(`${__dirname}/../__data__/data_old.json`)
+      ),
+      feed_items: JSON.parse(
+        fs.readFileSync(`${__dirname}/../__data__/feed_items.json`)
+      ),
+      data_updated: new Date()
+    })
+  });
+
+  it('should create a feed.xml file', () => {
```
blazeu: But it's not creating any file right?

it's returning an object containing the feed?

blazeu (changes requested):

tests/lib/rss.test.js
```diff
@@ -0,0 +1,59 @@
...
+ * RSS Library Test
+ */
+
+const fs = require('fs')
+const rss = require('../../lib/rss')
+
+describe('lib.rss', () => {
+  let rssResponse;
+
+  beforeAll(async () => {
```
blazeu: This will run once.

Each test should get fresh object to work with.

lib/rss.js
```diff
@@ -0,0 +1,191 @@
...
+  }, {})
+}
+
+module.exports = async ({
+  data: newData,
+  data_old: oldData,
+  feed_items: current,
+  data_updated: dataUpdated,
+}) => {
+  return new Promise(resolve => {
```
blazeu: Is there any reason for using Promise?

It's an async function already, should return Promise all the time.

But really, you don't need to be asynchronous in this case and it shouldn't return Promise ..
I don't see any method calls that's asynchronous in this function.

tests/lib/rss.test.js
andrewda: Sure I can do that. Why should we do this, though? The object isn't manipulated at all and the data should be the same.

lib/rss.js
andrewda: Yep, you're right. I mistakingly thought we were ending/writing inside a callback.

jayvdb:
Please squash your rss commits.  There is no need to create a file in one commit and then modify it in another , within one PR.  It is one code addition.

tests/lib/rss.test.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * RSS Library Test
+ */
+
+const fs = require('fs')
+const rss = require('../../lib/rss')
+
+describe('lib.rss', () => {
+  let rssResponse;
```
blazeu: ;

blazeu:
Tests are failing https://app.netlify.com/sites/gci-leaders/deploys/5a3ca2074c4b935f5955dcb0

tests/lib/rss.test.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * RSS Library Test
+ */
+
+const fs = require('fs')
+const rss = require('../../lib/rss')
+const mockdate = require('mockdate')
+
+mockdate.set('2017-12-22T08:05:39.000Z')
```
blazeu: Make sure this mock only applies inside the test that needs it and cleaned/removed after that.

tests/lib/rss.test.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+const fs = require('fs')
+const rss = require('../../lib/rss')
+const mockdate = require('mockdate')
+
+describe('lib.rss', () => {
+  let rssResponse
+
+  beforeEach(() => {
```
blazeu, line 12: Sure if the object will not be manipulated then use `beforeAll`, but we will definitely manipulate it in the future.

tests/lib/rss.test.js
```diff
@@ -0,0 +1,43 @@
...
+      ),
+      data_old: JSON.parse(
+        fs.readFileSync(`${__dirname}/../__data__/data_old.json`)
+      ),
+      data_updated: new Date('2017-12-22T08:44:01.327Z'),
+    })
+  })
+
+  afterAll(() => {
+    mockdate.reset()
```
blazeu: If you still use `beforeEach`, then change this to `afterEach`.
It doesn't make sense to mock the date everytime but only reset it once.

blazeu (commented):
Travis fails ..

tests/lib/rss.test.js
andrewda: Alright, left this as-is and changed `afterAll` to `afterEach` so they match.

lib/rss.js
```diff
@@ -0,0 +1,190 @@
...
+          ],
+        })
+      })
+
+      const feed = new RSS({
+        title: 'Google Code-in Leaders',
+        description: 'A feed for Google Code-in updates',
+        site_url: process.env.URL || 'https://gci-leaders.netlify.com',
+        feed_url: `${process.env.URL ||
+          'https://gci-leaders.netlify.com'}/build.xml`,
```
jayvdb: why `build.xml`?  It doesnt exist...?

andrewda: Good catch. Should be `feed.xml`. Fixed in latest push.

blazeu (approved):

Issue #97, opened 2017-12-20
Manually fix Github usernames?


ghost:
Hey, I'm one of the current leaders for MetaBrainz. My github account is linked as "eshansingh", but I recently changed my username to "naiveaiguy". Is there a way I can manually correct this in a PR?

blazeu:
"Manually", no.

But if you can find a way to "fix" or link the rest of non linked participants automatically, sure.

This is how our "guessing" of GitHub username currently work https://github.com/coala/gci-leaders/pull/62 & https://github.com/coala/gci-leaders/pull/74 & https://github.com/coala/gci-leaders/pull/75

There was a issue made because of you specifically https://github.com/coala/gci-leaders/issues/93
but the patch did not work expectedly to remove your old GitHub link.

blazeu:
As per https://github.com/coala/gci-leaders/pull/75

I think you just need to set your organization as public under your profile.

Might need some time because https://github.com/coala/gci-leaders/issues/93 is not fully fixed yet ;)

ghost:
Oh, I'm terribly sorry if I caused trouble >.< For MetaBrainz you don't actually join the org, so I can't set it as public. I'll close this for now.

ghost:
It appears my GitHub is now gone. Looking through the code right now to see if I can fix this.

andrewda:
@naiveaiguy Yea, there's no way to find it at the moment. The fix username caching PR was just to remove your 404ing username. We'll need to implement new methods to find your new username. See #100.

PR #99, opened 2017-12-20
Fix username caching

andrewda:
This will fix an issue where GitHub accounts are not updated
because they do not have a github_updated value.

Related to https://github.com/coala/gci-leaders/issues/93, https://github.com/coala/gci-leaders/issues/97

### DIFF
--- a/lib/scrape.js
+++ b/lib/scrape.js
@@ -276,7 +276,7 @@ function checkGitHubUserCacheExpired(user) {
 }
 
 async function freshenUserGitHubCache(user, existingUser, organization) {
-  if (!existingUser) {
+  if (!(existingUser && existingUser.github_updated)) {
     return {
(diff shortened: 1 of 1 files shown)


lib/scrape.js
```diff
@@ -276,7 +276,7 @@ function checkGitHubUserCacheExpired(user) {
 }
 
 async function freshenUserGitHubCache(user, existingUser, organization) {
-  if (!existingUser) {
+  if (!existingUser || !existingUser.github_updated) {
```
ghost: Minor change - I feel like this would be more elegant with `!(existingUser && existingUser.github_updated)`.

andrewda: Sure, works for me.

blazeu (approved):

Issue #102, opened 2017-12-21
Add StackOverflow links


jayvdb:
Use the Wikidata information to include StackOverflow links

nalinbhardwaj:
Claiming.

jayvdb:
Note that tags are site specific.

https://stackoverflow.com/questions/tagged/coala is valid
https://superuser.com/questions/tagged/coala is not

PR #109, opened 2017-12-22
scrape.js,main.html: Add SE links

nalinbhardwaj:
Adds Stack Exchange links for orgs. Chooses primary in case of multiple. Dynamically chooses logo to be displayed.
Since stack exchange themselves uses a common CDN for their websites, we don't have to worry about it.

Closes https://github.com/coala/gci-leaders/issues/102

### DIFF
--- a/lib/scrape.js
+++ b/lib/scrape.js
@@ -4,6 +4,7 @@ const fs = require('fs')
 const json2yaml = require('json2yaml')
 const validUsername = require('valid-github-username')
 const wdk = require('wikidata-sdk')
+const cheerio = require('cheerio')
 
 const GH_BASE = 'https://github.com'
 const GH_USER_BASE = `${GH_BASE}/users`
@@ -15,6 +16,16 @@ const MIN_SEARCH_SCORE = 10
 
 // The time to cache GitHub usernames for in milliseconds
 const GITHUB_CACHE_TIME = 2 * 24 * 60 * 60 * 1000
+
+// P1482 is the Wikidata property for Stack Exchange tag
+const WIKI_SE_TAG = 'P1482'
+
+// Prioritisation order of ranks on Wikidata, smaller is better
+const WIKI_CLAIM_PRIORITY = {
+  preferred: 1,
+  normal: 2,
+  deprecated: 3,
+}
 
 const CHAT_IMAGES = {
   GITTER: 'images/logos/gitter.png',
@@ -184,6 +195,38 @@ async function findOrganization({
   return null
 }
 
+async function fetchSEtag(wikidata_id) {
+  const wikiAPIurl = wdk.getEntities({
+    ids: [wikidata_id],
+    props: ['claims'],
+  })
+  const page = await fetch(wikiAPIurl)
+  let data = await page.json()
+  data = data.entities[wikidata_id]
+
+  if (!(WIKI_SE_TAG in data.claims)) {
+    return null
+  }
+
+  data.claims[WIKI_SE_TAG].sort(
+    (a, b) => WIKI_CLAIM_PRIORITY[a.rank] - WIKI_CLAIM_PRIORITY[b.rank]
+  )
+
+  const url = wdk.simplify.claim(data.claims[WIKI_SE_TAG][0])
+  const response = await fetch(url)
+  const txt = await response.text()
+  const html = cheerio.load(txt)
+
+  const image = html('link')
+    .filter((i, el) => html(el).attr('rel') === 'apple-touch-icon image_src')
+    .attr('href')
+
+  return {
+    url,
+    image,
+  }
+}
+
 async function findWiki(name) {
   const APIurl = wdk.searchEntities({
     search: name,
@@ -194,6 +237,8 @@ async function findWiki(name) {
   if (response && response.search.length) {
     const wikidataurl = response.search[0].url
     const wikidataid = response.search[0].id
+
+    const stack_exchange = await fetchSEtag(wikidataid)
 
     const entityAPIurl = wdk.getEntities({
       ids: [wikidataid],
@@ -210,11 +255,13 @@ async function findWiki(name) {
         }))
       )
       return {
+        stack_exchange,
         wikipedia: wikipediaurls,
         wikidata: wikidataurl,
       }
     } else {
       return {
+        stack_exchange,
         wikipedia: null,
         wikidata: wikidataurl,
       }
@@ -223,6 +270,7 @@ async function findWiki(name) {
     return {
       wikipedia: null,
       wikidata: null,
+      stack_exchange: null,
     }
   }
 }
@@ -357,6 +405,7 @@ async function fetchOrgsWithData() {
       },
       wikipedia_urls: orgWiki[index].wikipedia,
       wikidata_url: orgWiki[index].wikidata,
+      stack_exchange: orgWiki[index].stack_exchange,
     })
   })
 
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   "license": "MIT",
   "dependencies": {
     "chattie": "0.0.4",
+    "cheerio": "^1.0.0-rc.2",
     "feed-read-parser": "^0.0.6",
     "find-rss": "^1.6.4",
     "jquery": "^3.2.1",
--- a/templates/main.html
+++ b/templates/main.html
@@ -91,6 +91,11 @@
                 <img src="images/logos/wikidata.png" class="chat" alt="See them on Wikidata" />
               </a>
             {{/wikidata_url}}
+            {{#stack_exchange}}
+              <a href="{{stack_exchange.url}}">
+                <img src="{{stack_exchange.image}}" class="chat" alt="Do Q&amp;A on Stack Exchange" />
+              </a>
+            {{/stack_exchange}}
             <div class="org-leaderboard">
               <ul>
                 {{#leaders}}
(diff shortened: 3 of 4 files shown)


templates/main.html
```diff
@@ -91,6 +91,11 @@ <h3>
                 <img src="images/logos/wikidata.png" class="chat" alt="See them on Wikidata" />
               </a>
             {{/wikidata_url}}
+            {{#stack_exchange}}
+              <a href="{{stack_exchange.url}}">
+                <img src="{{stack_exchange.image}}" class="chat" alt="Do Q&A on Stack Exchange" />
```
wisn: Use `&amp;` instead of `&`.

templates/main.html
```diff
@@ -172,6 +177,11 @@ <h3>
...
```
wisn: Use `&amp;` instead of `&`.

lib/scrape.js
```diff
@@ -184,6 +185,33 @@ async function findOrganization({
...
+  let data = await page.json()
+  data = data.entities[wikidata_id]
+  //P1482 is Stack Exchange tag
+  if ('P1482' in data.claims) {
+    const url = wdk.simplify.claim(data.claims.P1482[0])
+    const response = await fetch(url)
+    const html = cheerio.load(await response.text())
+
+    const img = html('link')
+      .filter(function() {
```
wisn: Read about <https://github.com/airbnb/javascript#arrow-functions> please.

lib/scrape.js
```diff
@@ -184,6 +185,33 @@ async function findOrganization({
   return null
 }
 
+async function fetchSEtag(wikidata_id) {
+  const url = wdk.getEntities({
+    ids: [wikidata_id],
+    props: ['claims'],
+  })
+  const page = await fetch(url)
+  let data = await page.json()
```
wisn: Why use `await` for `page.json()`?

lib/scrape.js
```diff
@@ -184,6 +185,33 @@ async function findOrganization({
...
+  data = data.entities[wikidata_id]
+  //P1482 is Stack Exchange tag
+  if ('P1482' in data.claims) {
+    const url = wdk.simplify.claim(data.claims.P1482[0])
+    const response = await fetch(url)
+    const html = cheerio.load(await response.text())
```
wisn: Better place that `await` in a variable first before apply it to a function argument.

lib/scrape.js
```diff
@@ -184,6 +185,33 @@ async function findOrganization({
...
+
+    const img = html('link')
+      .filter(function() {
+        return html(this).attr('rel') === 'apple-touch-icon image_src'
+      })
+      .attr('href')
+    return {
+      url: url,
```
wisn: You can directly use something like this

```javascript
return {
  url,
 // ...
```

It is the same as `url: url`.

lib/scrape.js
```diff
@@ -212,17 +240,20 @@ async function findWiki(name) {
       return {
         wikipedia: wikipediaurls,
         wikidata: wikidataurl,
+        stack_exchange: await fetchSEtag(wikidataid),
```
wisn: Call that `await` in a variable before that `const APIurl`.

lib/scrape.js
```diff
@@ -212,17 +240,20 @@ async function findWiki(name) {
...
       }
     } else {
       return {
         wikipedia: null,
         wikidata: wikidataurl,
+        stack_exchange: await fetchSEtag(wikidataid),
```
wisn: Call that `await` in a variable before that `const APIurl`.

lib/scrape.js
nalinbhardwaj: Because it returns a promise?
https://developer.mozilla.org/en-US/docs/Web/API/Body/json

lib/scrape.js
nalinbhardwaj: Yeah, I added use of ‘this’ for this purpose but forgot to do it 😛. Will fix, thanks.

package.json
```diff
@@ -29,7 +29,8 @@
     "mustache": "^2.3.0",
     "node-fetch": "^1.7.3",
     "valid-github-username": "0.0.1",
-    "wikidata-sdk": "^5.5.0"
+    "wikidata-sdk": "^5.5.0",
+    "cheerio": "^1.0.0-rc.2"
```
blazeu: Sort it.

Use `npm install`

lib/scrape.js
```diff
@@ -184,6 +185,32 @@ async function findOrganization({
...
+async function fetchSEtag(wikidata_id) {
+  const url = wdk.getEntities({
+    ids: [wikidata_id],
+    props: ['claims'],
+  })
+  const page = await fetch(url)
+  let data = await page.json()
+  data = data.entities[wikidata_id]
+  //P1482 is Stack Exchange tag
+  if ('P1482' in data.claims) {
```
blazeu: P1482 should be in a constant.

blazeu (dismissed):

lib/scrape.js
```diff
@@ -184,6 +185,32 @@ async function findOrganization({
...
+    const url = wdk.simplify.claim(data.claims.P1482[0])
+    const response = await fetch(url)
+    const txt = await response.text()
+    const html = cheerio.load(txt)
```
blazeu: Is there any API for SE? We could use that instead of dirty scrapping.

nalinbhardwaj: There is, but I don't think it's very useful since there is no direct way to get the favicon, and besides, it'll need a key etc. Link: https://api.stackexchange.com/docs

lib/scrape.js
```diff
@@ -184,6 +188,32 @@ async function findOrganization({
...
+  if (WIKI_SE_TAG in data.claims) {
+    const url = wdk.simplify.claim(data.claims[WIKI_SE_TAG][0])
+    const response = await fetch(url)
+    const txt = await response.text()
+    const html = cheerio.load(txt)
+
+    const image = html('link')
+      .filter((i, el) => html(el).attr('rel') === 'apple-touch-icon image_src')
+      .attr('href')
+    return {
```
blazeu: Blank line before this return; for readability.

lib/scrape.js
wisn: Well, it just weird to use await twice for grabbing one data you need. Anyway, have you read this?

> The Promise returned from fetch() won’t reject on HTTP error status even if the response is an HTTP 404 or 500. Instead, it will resolve normally (with ok status set to false), and it will only reject on network failure or if anything prevented the request from completing.

Well, WikiData rarely encounter 500 error.

lib/scrape.js
```diff
@@ -184,6 +188,32 @@ async function findOrganization({
...
+async function fetchSEtag(wikidata_id) {
+  const url = wdk.getEntities({
+    ids: [wikidata_id],
+    props: ['claims'],
+  })
+  const page = await fetch(url)
+  let data = await page.json()
+  data = data.entities[wikidata_id]
+  //P1482 is Stack Exchange tag
+  if (WIKI_SE_TAG in data.claims) {
```
blazeu: Invert this and you'll decrease the indent for the long code below.

```
if (!
  return
```

lib/scrape.js
nalinbhardwaj: Wikipedia API suggests that the encountering is negligible? And even if it does encounter that won't we just end up with a null object then? I don't think we have handled for that edge case in the other parsings as well..

blazeu: @wisn I did use `node-fetch` when I started this project which has native Fetch-like api thus needing that 2 await.

There's a benefit if we run this on the client i.e. same API (isomorphic; easily switch to [isomorphic-fetch](https://github.com/matthew-andrews/isomorphic-fetch)). But we don't.

Bad choice really, we should make a task for switching to something else, e.g. [axios](https://github.com/axios/axios)

wisn: @blazeu I see. Yes, we need something else. I up vote axios.

wisn: @nalinbhardwaj

> And even if it does encounter that won't we just end up with a null object then

Yes, but, a null object doesn't have `entities` attribute. I just thinking about possibility.

lib/scrape.js
```diff
@@ -184,6 +188,33 @@ async function findOrganization({
...
 
+async function fetchSEtag(wikidata_id) {
+  const wikiAPIurl = wdk.getEntities({
+    ids: [wikidata_id],
+    props: ['claims'],
+  })
+  const page = await fetch(wikiAPIurl)
+  let data = await page.json()
+  data = data.entities[wikidata_id]
+  //P1482 is Stack Exchange tag
```
blazeu: No need for this comment.

wisn: Can't you use a better comment style?

nalinbhardwaj: Can you elaborate? I moved the comment to [here](https://github.com/coala/gci-leaders/pull/109/files#diff-e7f5b3f014790acdf91e958a37dc5f2cR20) with a little more detail.

blazeu: Redundant comment, it's a constant at the top, already has comment.

lib/scrape.js
```diff
@@ -212,17 +246,20 @@ async function findWiki(name) {
       return {
         wikipedia: wikipediaurls,
         wikidata: wikidataurl,
+        stack_exchange,
```
wisn: Please read this one <https://github.com/airbnb/javascript#objects--grouped-shorthand>

nalinbhardwaj: Fixed, sorry for so many style errors.

wisn:
All good I think.

wisn (approved):

jayvdb (changes requested):
Commit message is misleading, where 'preferred' is actually 'first in the array'.  They are not equivalent.

https://www.wikidata.org/wiki/Q46507419 contains lots of jboss tags on all the SE sites. ;-)

Wikidata has a concept of 'preferred rank', which I have set on the 'gis' SE site.

However it isnt shown by this patch.  Instead this patch shows the 'first', which is SO.

Either change the commit message and create an issue to use preferred, or update the code to use preferred.

jayvdb (approved):

Issue #111, opened 2017-12-23
Use GraphQL API


andrewda:
We should use GitHub's GraphQL API instead of the the REST API. This will let us get a lot more data in many fewer requests.

You should preferably already have some GraphQL experience in order to properly make use of everything GraphQL has to offer, such as pagination and query variables.

Check out https://www.npmjs.com/package/graphql-client for making requests.

li-boxuan:
@andrewda @jayvdb One problem with `searchGitHubOrgs` function: via REST API, GitHub would return a `score` to represent matching degree. 

[scrape.js#L287-L289](https://github.com/coala/gci-leaders/blob/master/lib/scrape.js#L287-L289) uses `score` attribute to see if it reaches threshold.

However, as far as I am concerned, GraphQL API would not return `score` or anything like that.
```graphql
query($name: String!) {
    search(type:USER, query:$name, first:1) {
        nodes {
            ...on Organization {
                login
            }
        }
    }
}
```
I would like to work on this issue.

li-boxuan:
Another thing is, the function `findGitHubUserInOrg` on [scrape.js#L505](https://github.com/coala/gci-leaders/blob/master/lib/scrape.js#L505) uses `${GH_ORG_BASE}/${org}/people?query=${user}` and regex to check if a user is in an org. This is ad-hoc but doesn't need any API quota. REST API and GraphQL API can do this check but of course, would use API quota. Is this design by intention?

jayvdb:
Regarding `findGitHubUserInOrg`, the choice of what to use is made based on how to do as much validation without hitting the limit.

PR #114, opened 2017-12-23
Identify GCI users in repository information

### LINKED ISSUE #100: Find users in repository information
We should check the `watchers`, `stargazers` and `forks` for all repositories owned by an organization to see if the user appears there. Luckily, GraphQL makes this _super_ easy.

```GraphQL
{
  organization(login: "coala") {
    repositories(first: 100) {
      nodes {
        watchers(first: 100) {
          nodes {
            login
            name
          }
        }
        stargazers(last: 100) {
          nodes {
            login
            name
          }
        }
        forks(first: 100) {
          nodes {
            owner {
              login
            }
          }
        }
      }
    }
  }
}
```

Check all those `login`s and `name`s against the GCI `display_names` in the organization to find matches. I did a manual check and this will find _at least_ another 5-10 matches, including 3 more from coala.

andrewda:
This begins to identify users in several meta repository stats,
including stargazers, watchers and forks.

Closes #100

***

A few notes:
- (fixed in ed4f74e47601394f7f9e3281613cf05c7272b380) ~~The number of users might not be higher in the build until the cache expires. This is a separate issue: we should bypass caching stuff if there is no `github_account` property on the user and jump straight to the checking stuff.~~
- We don't check the time of the watch/star events in order to see if it happened during GCI (as @jayvdb wanted in https://github.com/coala/gci-leaders/issues/100#issuecomment-353432948) because that information isn't returned to us through the API.
- As far as I can tell, we are unable to access the name of the fork owner (GraphQL type: `RepositoryOwner`) without making another request (to get the `User` from the `RepositoryOwner`). That would involve several hundred more requests and probably being rate limited by GitHub, unless someone can figure out a way to do it all in one query – I wasn't able to after a few hours of trial and error.
- This vastly reduces the overall number of API calls we have to make. Instead of calling `findGitHubUserInOrg` for every user, we only have to search pre-fetched information for a large number of users.
- This sets the base for #111, meaning it'll be a low difficulty issue for anyone with GraphQL experience, and a medium difficulty issue for anyone without experience.
- There are a lot of `package-lock.json` changes that get fixed in #95 – this definitely should not have 900+ deletions.

### DIFF
--- a/lib/queries/index.js
+++ b/lib/queries/index.js
@@ -1,0 +1,3 @@
+const { loadQuery } = require('../utils')
+
+module.exports.GITHUB_REPO_INFO_QUERY = loadQuery('github_repo_info')
--- a/lib/scrape.js
+++ b/lib/scrape.js
@@ -1,4 +1,5 @@
 const fetch = require('node-fetch')
+const GraphQL = require('graphql-client')
 const chattie = require('chattie')
 const fs = require('fs')
 const json2yaml = require('json2yaml')
@@ -7,10 +8,14 @@ const validUsername = require('valid-github-username')
 const wdk = require('wikidata-sdk')
 const cheerio = require('cheerio')
 
+const { GITHUB_REPO_INFO_QUERY } = require('./queries')
+const { getLatestCommitMessage } = require('./utils')
+
 const GH_BASE = 'https://github.com'
 const GH_USER_BASE = `${GH_BASE}/users`
 const GH_ORG_BASE = `${GH_BASE}/orgs`
 const GH_API_BASE = 'https://api.github.com'
+const GH_GQL_BASE = 'https://api.github.com/graphql'
 const GCI_API_BASE = 'https://codein.withgoogle.com/api'
 
 const MIN_SEARCH_SCORE = 10
@@ -54,6 +59,13 @@ const GH_API_OPTIONS = {
     : {},
 }
 
+const GH_GQL_OPTIONS = {
+  url: GH_GQL_BASE,
+  headers: process.env.GITHUB_TOKEN
+    ? { Authorization: `bearer ${process.env.GITHUB_TOKEN}` }
+    : {},
+}
+
 const GH_WEB_OPTIONS = {
   headers: {
     Accept: 'text/html',
@@ -64,6 +76,11 @@ const GH_WEB_OPTIONS = {
   },
   compress: false,
 }
+
+const client = GraphQL(GH_GQL_OPTIONS)
+
+let COMPETITION_OPEN
+let BUST_GITHUB_CACHE
 
 let existingData = []
 try {
@@ -94,6 +111,71 @@ async function fetchLeaders(id) {
   const res = await fetch(`${GCI_API_BASE}/program/current/organization/${id}`)
   const { leaders } = await res.json()
   return leaders
+}
+
+let repositoryInfo = {}
+async function fetchRepositoryInfo(org) {
+  if (repositoryInfo[org]) return repositoryInfo[org]
+
+  const { data } = await client.query(GITHUB_REPO_INFO_QUERY, { org })
+
+  if (data) {
+    const info = data.organization.repositories.nodes.map(node => ({
+      watchers: node.watchers.nodes,
+      stargazers: node.stargazers.nodes,
+      forks: node.forks.nodes,
+    }))
+
+    repositoryInfo[org] = info
+
+    return info
+  } else {
+    return []
+  }
+}
+
+async function getGitHubUserFromRepoInfo(org, displayName, shortName) {
+  let repos = []
+  try {
+    repos = await fetchRepositoryInfo(org)
+  } catch (e) {
+    console.error(`Could not fetch repository info for ${org}...`)
+  }
+
+  let logins = []
+  let names = {}
+
+  repos.forEach(repo => {
+    logins = logins
+      .concat(repo.watchers.map(u => u.login.toLowerCase()))
+      .concat(repo.stargazers.map(u => u.login.toLowerCase()))
+      .concat(
+        repo.forks
+          .map(u => {
+            const createdAt = new Date(u.createdAt)
+            if (createdAt.getTime() > COMPETITION_OPEN.getTime()) {
+              return u.owner.login.toLowerCase()
+            }
+          })
+          .filter(login => login)
+      )
+
+    repo.watchers.forEach(watcher => {
+      if (watcher.name) {
+        names[watcher.name.toLowerCase()] = watcher.login
+      }
+    })
+  })
+
+  logins = logins.filter((item, pos, self) => self.indexOf(item) === pos)
+
+  if (logins.includes(shortName.toLowerCase())) {
+    return shortName
+  }
+
+  if (names[displayName.toLowerCase()]) {
+    return names[displayName.toLowerCase()]
+  }
 }
 
 async function checkGitHubUserExists(user) {
@@ -366,25 +448,43 @@ async function findGitHubUser(displayName, org) {
 
   const shortName = validUsername(displayName)
 
-  const username = await findGitHubUserInOrg(displayName, org)
-  if (username) return username
+  let userInOrg
+  try {
+    userInOrg = await findGitHubUserInOrg(displayName, org)
+  } catch (e) {
+    console.error(`Failed to find user ${displayName} in org ${org}...`)
+  }
+  if (userInOrg) {
+    console.log(`${displayName}: ${userInOrg} (method: user in org)`)
+    return userInOrg
+  }
 
   let user
   try {
     user = await getGitHubUser(shortName)
   } catch (e) {
-    console.error(`Got error when finding user ${shortName}...`)
+    console.error(`Failed to find user ${shortName}...`)
   }
-  if (!user) return
+
+  if (!user) {
+    const userFromRepo = await getGitHubUserFromRepoInfo(
+      org,
+      displayName,
+      shortName
+    )
+
+    if (!userFromRepo) {
+      return
+    }
+
+    user = userFromRepo
+  }
 
   const login = user.login
 
-  const { competition_open_starts } = await fetchProgram()
-
   const updatedTime = new Date(user.updated_at)
-  const openTime = new Date(competition_open_starts)
 
-  if (updatedTime.getTime() - openTime.getTime() < 0) return
+  if (updatedTime.getTime() - COMPETITION_OPEN.getTime() < 0) return
 
   let orgs = []
   try {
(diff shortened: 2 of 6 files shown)


lib/queries/index.js
```diff
@@ -0,0 +1,30 @@
+module.exports.REPO_INFO_QUERY = `
```
jayvdb: store in a `repo_info.graphql` ?
the variables are part of the language.

lib/scrape.js
```diff
@@ -69,6 +84,71 @@ async function fetchLeaders(id) {
...
+
+  let logins = []
+  let names = {}
+
+  repos.forEach(repo => {
+    logins = logins
+      .concat(repo.watchers.map(u => u.login.toLowerCase()))
+      .concat(repo.stargazers.map(u => u.login.toLowerCase()))
+      .concat(
+        repo.forks
```
jayvdb, line 153: so I could create account https://github.com/WinningIt , fork a Ubuntu repo, and my fake account gets linked?

lib/scrape.js
```diff
@@ -232,25 +312,39 @@ async function findGitHubUser(displayName, org) {
...
   const shortName = validUsername(displayName)
 
-  const username = await findGitHubUserInOrg(displayName, org)
-  if (username) return username
+  const userFromRepo = await getGitHubUserFromRepoInfo(
+    org,
+    displayName,
+    shortName
+  )
+  if (userFromRepo) return userFromRepo
```
jayvdb: so I could create account https://github.com/WinningIt , fork a Ubuntu repo, and my fake account gets linked?

jayvdb (changes requested):

lib/scrape.js
andrewda: I suppose you could if you wanted to be malicious, but with the current implementation you could create that same fake account and make a BS issue on any Ubuntu repository and the account would be linked. Or even just do a PR review.

jayvdb:
Add a commit to forcibly clear the cache of unlinked usernames so we can see it working.  We'll no doubt need to redo this occasionally.

lib/queries/index.js
```diff
@@ -0,0 +1,5 @@
+const fs = require('fs')
+
+module.exports.REPO_INFO_QUERY = fs
+  .readFileSync(`${__dirname}/repo_info.graphql`)
+  .toString()
```
blazeu: We'll be doing this very often, might as well just create a function to load `.graphql`

Probably put that in `utils.js` file or there might be an existing npm packages to do it.

andrewda: Sounds good. It'll be like 3 lines of code to do it ourselves so may as well keep it light weight.

lib/queries/index.js
```diff
@@ -0,0 +1,3 @@
+const { loadQuery } = require('../utils')
```
wisn, line 1: GCI Leaders now using Babel, isn't it? Use `import` instead?

lib/queries/index.js
```diff
@@ -0,0 +1,3 @@
...
+
+module.exports.REPO_INFO_QUERY = loadQuery('repo_info')
```
wisn: Better to use

```javascript
export const REPO_INFO_QUERY = // ...
```

or

```javascript
export {
  REPO_INFO_QUERY: // ...
}
```

I think

lib/scrape.js
```diff
@@ -1,16 +1,22 @@
...
+const { REPO_INFO_QUERY } = require('./queries')
+
 const GH_BASE = 'https://github.com'
 const GH_USER_BASE = `${GH_BASE}/users`
 const GH_ORG_BASE = `${GH_BASE}/orgs`
 const GH_API_BASE = 'https://api.github.com'
+const GH_GQL_BASE = 'https://api.github.com/graphql'
 const GCI_API_BASE = 'https://codein.withgoogle.com/api'
 
+let COMPETITION_OPEN
```
wisn: Group that `let` and `const`. See <https://github.com/airbnb/javascript#variables--const-let-group>.

lib/scrape.js
```diff
@@ -69,6 +84,71 @@ async function fetchLeaders(id) {
...
+      )
+
+    repo.watchers.forEach(watcher => {
+      if (watcher.name) {
+        names[watcher.name.toLowerCase()] = watcher.login
+      }
+    })
+  })
+
+  logins = logins.filter((item, pos, self) => self.indexOf(item) == pos)
```
wisn: Better to use `===` instead of `==`. See <https://github.com/airbnb/javascript#comparison--eqeqeq>.

lib/scrape.js
```diff
@@ -276,15 +370,26 @@ function checkGitHubUserCacheExpired(user) {
 }
 
 async function freshenUserGitHubCache(user, existingUser, organization) {
-  if (!(existingUser && existingUser.github_updated)) {
+  if (
+    !(
+      existingUser &&
+      existingUser.github_updated &&
```
wisn, line 528: See <https://github.com/airbnb/javascript#control-statements>.

lib/utils.js
```diff
@@ -0,0 +1,4 @@
+const fs = require('fs')
```
wisn, line 1: Use `import` instead?

lib/utils.js
```diff
@@ -0,0 +1,4 @@
...
+
+module.exports.loadQuery = name =>
```
wisn: Use `export const loadQuery` instead.

wisn (changes requested):

lib/scrape.js
jayvdb: An issue is a very noticable action.  It is a new thing, and the malicious person is the creator.
A review is not quite as noticeable, but it would need to be done on an obscure open PR in order to not be noticed.
Either would likely be seen as intentional confusion, maybe suitable for GitHub to investigate as 'abuse'.

stars/watches/forks are almost invisible.  And it is very hard to file 'abuse' with GitHub for doing those actions.

Your issue is a specific attempt to do https://github.com/coala/gci-leaders/issues/8 .

The biggest problem is that so far this algorithm has a lower confidence level than the previous algorithms, but it is being used first.  This could be used to override the existing algorithms, which are more reliable.

Once you have a potential match, which wasnt found using the existing more reliable algorithms, you need to look at the match profiles to determine how you can increase the confidence level of your match.  The more effort you require of the abuser, the more likely their abuse can only be viewed as intentional or at least highly suspicious.

Probably also a good idea to annotate each match with the matching method used.

API hits isn't relevant now, as the hits can grow over time.  You can include probable matches in the yaml which are not included in the rendered page, as they are needing more analysis deferred until a subsequent build has extra API calls to use.

lib/scrape.js
andrewda: I agree with Airbnb, but this is a styling thing in our ESLint config that we should probably change in another PR.

lib/queries/index.js
andrewda: Only frontend code (`/static`) uses Babel at the moment, so this isn't yet possible.

lib/queries/index.js
andrewda: 👍 I'll make a PR after this to setup Babel for the backend and start using ES2017 syntax.

lib/scrape.js
andrewda: How about running the same checks we do on the user found this way as we do for users found by removing spaces in the display name? i.e. we make sure the user has at least a commit, issue, PR or review on a repository of the org

That way we can have the same confidence in this prediction as we do for all the others

lib/queries/index.js
wisn: Alright.

wisn (approved):

lib/scrape.js
andrewda: The problem with ^ is that it cuts the number of matches in half and gets rid of some valid matches, but it does result in a much more accurate result.

lib/scrape.js
```diff
@@ -1,21 +1,26 @@
...
 const GH_ORG_BASE = `${GH_BASE}/orgs`
 const GH_API_BASE = 'https://api.github.com'
+const GH_GQL_BASE = 'https://api.github.com/graphql'
 const GCI_API_BASE = 'https://codein.withgoogle.com/api'
 
 const MIN_SEARCH_SCORE = 10
 
 // The time to cache GitHub usernames for in milliseconds
 const GITHUB_CACHE_TIME = 2 * 24 * 60 * 60 * 1000
+const BUST_GITHUB_CACHE = true
```
jayvdb: If I understand correctly, this will keep busting the cache until someone sets this to false.

Each cache bust should be a once-off operation which automatically reverts to non-cache busting on the next build.
There are ways to use of elements of the merge that can be used to detect whether the bust was complete.
Or store the cache bust timestamp in the deploy, and dont let it occur again within a short period of time.

lib/scrape.js
jayvdb: Is the total number of matches increased?
If so, loosing some valid matches is ok.

andrewda: I believe it only increases by 1 if we verify using our current method, but will increase by 3 if we allow contributions to any of the GCI orgs to count (not just the one they are a leader for).

lib/scrape.js
andrewda: Would using Travis to check the commit name for a specific keyword work?

andrewda: Actually netlify doesn't expose the commit message, so we're going to have to check with `git log -1 --pretty=%B`. Another alternative is an env variable so that the cache busting can't be abused by someone other than an admin?

lib/scrape.js
jayvdb: 1 is enough to meet the requirements of the task

jayvdb:
`repo_info.graphql` should be renamed to include `github` as that is what it is. GitLab doesnt support it yet I believe, due to [patent issues](https://www.theregister.co.uk/2017/09/20/gitlab_suspends_graphql_project_over_facebook_license_terms/) , but it could, or other uses of GraphQL could arise, and it doesnt hurt to be explicit in the filename.

jayvdb:
@gitmate-bot fastforward

jayvdb (approved):

jayvdb:
ack https://github.com/coala/gci-leaders/pull/114/commits/6a17aac454c61cfc7f0dafb2e7735270812e70ba https://github.com/coala/gci-leaders/pull/114/commits/97acdb848ff554ead7eb094fc4395e86eec30e49 https://github.com/coala/gci-leaders/pull/114/commits/b33962739d17a13bf8d16ca1606015efa9c6eabb

jayvdb:
@gitmate-bot fastforward

Issue #9, opened 2017-12-05
Link to GitLab organisation for each Google Code-in organisation 


jayvdb:
Link to the GitLab org for each Google Code-in organisation, wherever possible, ensuring that only highly probable links are made.

Use a GitLab logo for the link.

jayvdb:
[GitMate.io](https://gitmate.io) thinks the label prediction failed because there are not enough labels. Three labels must have been set on at least three issues each.

The contributor prediction failed because there are not enough contributors. Three contributors must have authored comments on at least three issues each.

PR #117, opened 2017-12-25
Link organizations to their GitLab accounts

andrewda:
This links an organization with their GitLab account by finding it
through their Wikidata page.

Closes https://github.com/coala/gci-leaders/issues/9

Currently only LibreHealth has their GitLab account linked on Wikidata (see https://www.wikidata.org/wiki/Q45341906). More orgs' GitLab accounts will be listed as they are linked on their respective Wikidata pages.


lib/scrape.js
```diff
@@ -233,6 +243,51 @@ async function fetchSEtag(wikidata_id) {
...
+    return
+  }
+
+  const instanceClaim = entity.claims[WIKI_INSTANCE_OF]
+
+  if (!instanceClaim) {
+    return
+  }
+
+  const instance = instanceClaim[0].mainsnak.datavalue.value.id
```
jayvdb: need to check all claims.
(or use sparql)

templates/main.html
```diff
@@ -57,6 +57,15 @@ <h3>
                 />
               </a>
             {{/github}}
+            {{#gitlab}}
+              <a href="https://gitlab.com/{{gitlab}}">
+                <img
+                  src="images/logos/gitlab.png"
```
jayvdb, line 63: If you are storing a copy in the repository, please ensure that the licensing of the artwork is complied with.

andrewda: It's under [CC BY-NC-SA 4.0](https://creativecommons.org/licenses/by-nc-sa/4.0/) so yea we have to provide attribution. Should I put this in the footer?

jayvdb: Storing it in the repo means it needs to be attributed in the relevant directory of the repo, as it isnt covered by the repo license.

jayvdb: a README in that directory would be useful.  Also for CC licenses, the full text of the license isnt necessary.  A link is acceptable, and a link to the license file in their repo is really good, as it allows the reader to find the original, which is the intent of CC.

andrewda: So instead of having the full text in the LICENSE file, just have a link to their license in the LICENSE file? And show which images are under which licenses in a README file?

lib/scrape.js
```diff
@@ -233,6 +243,53 @@ async function fetchSEtag(wikidata_id) {
...
+  const wikiAPIurl = wdk.getEntities(wikidata_id)
+  const page = await fetch(wikiAPIurl)
+  const { entities } = await page.json()
+  const entity = entities[wikidata_id]
+
+  if (!(WIKI_ACCOUNT_ON in entity.claims)) {
+    return
+  }
+
+  const gitlabClaim = entity.claims[WIKI_ACCOUNT_ON].find(
```
jayvdb: this will fail if there is more than one?

it is a good idea to sort the claims by rank ; see https://github.com/coala/gci-leaders/pull/117/files#diff-e7f5b3f014790acdf91e958a37dc5f2cL217
`data.claims[WIKI_SE_TAG].sort`

templates/main.html
```diff
@@ -195,6 +204,7 @@ <h3>
     </div>
     <footer>
       <small data-i18n="footer-text">Google Code-in and the Google Code-in logo are trademarks of Google Inc.</small>
+      <small>The GitLab logo is licensed under <a href="https://gitlab.com/gitlab-com/gitlab-artwork/blob/master/LICENSE">CC BY-NC-SA 4.0</a> by GitLab Inc.</small>
```
jayvdb: shouldnt this be i18n-able?

andrewda: Any idea how to use i18n with the link? If I include the link, it renders all the `<a href="...` as text on the page.

Maybe something like this would work, but it's not altogether pretty:

```
<small><span data-i18n="gitlab-copyright-1">The GitLab logo is licensed under </span><a href="https://gitlab.com/gitlab-com/gitlab-artwork/blob/master/LICENSE">CC BY-NC-SA 4.0</a><span data-i18n="gitlab-copyright-2"> by GitLab Inc.</span></small>

blazeu: ^ That's ok.

Also add `<br>` after Google copyright.

andrewda: @blazeu done 👍

blazeu (approved):

Issue #77, opened 2017-12-12
Import pre-2017 Finalists and winners


jayvdb:
It appears there is no raw data, so it will need to be scraped from the website.

Present on a separate page, either as one page , or a page per year.

blazeu:
I'd recommend pulling in some scrapping library such as [x-ray](https://github.com/matthewmueller/x-ray) or something else if there's a better one.

No dirty regex ;)

PR #118, opened 2017-12-27
Import pre-2017 Finalists and winners

tioirawan:
This import pre-2017 Finalists and winners to JSON file

Closes https://github.com/coala/gci-leaders/issues/77

### DIFF
--- a/.coafile
+++ b/.coafile
@@ -15,6 +15,7 @@ bears = LineLengthBear
 [all.links]
 bears = InvalidLinkBear
 link_ignore_regex = (github\.com|\{|\$)
+link_ignore_list = https://www.google-melange.com/**
 
 [js]
 files = lib/**/*.js, src/js/**/*.js
--- a/lib/main.js
+++ b/lib/main.js
@@ -2,6 +2,7 @@ const glob = require('glob')
 const fs = require('fs')
 
 const rss = require('./rss')
+const pre2017 = require('./pre2017')
 ;(async () => {
   const files = glob.sync(`${__dirname}/../out/*.json`)
   const fileContents = files.map(f => JSON.parse(fs.readFileSync(f).toString()))
@@ -17,4 +18,10 @@ const rss = require('./rss')
   const rssResult = rss(data)
   fs.writeFileSync(`${__dirname}/../out/feed_items.json`, rssResult.feed_items)
   fs.writeFileSync(`${__dirname}/../out/feed.xml`, rssResult.feed_xml)
+
+  const pre2017Result = await pre2017()
+  fs.writeFileSync(
+    `${__dirname}/../out/pre2017.json`,
+    JSON.stringify(pre2017Result, null, 2)
+  )
 })()
--- a/lib/pre2017.js
+++ b/lib/pre2017.js
@@ -1,0 +1,131 @@
+const fetch = require('node-fetch')
+const cheerio = require('cheerio')
+const https = require('https')
+
+const PRE_URL = [
+  'https://codein.withgoogle.com/archive/2016',
+  'https://codein.withgoogle.com/archive/2015',
+  'https://www.google-melange.com/archive/gci/2014',
+  'https://www.google-melange.com/archive/gci/2013',
+  'https://www.google-melange.com/archive/gci/2012',
+  'https://www.google-melange.com/archive/gci/2011',
+  'https://www.google-melange.com/archive/gci/2010',
+]
+
+const CODEIN_WIKI_URL = 'https://en.wikipedia.org/wiki/Google_Code-in'
+
+async function fetchText(url) {
+  const agent = new https.Agent({
+    rejectUnauthorized: false,
+  })
+  const res = await fetch(url, {
+    agent,
+  })
+
+  return await res.text()
+}
+
+async function fetchPre2017(year, url) {
+  const htmlResponse = await fetchText(url)
+  const json =
+    year > 2014
+      ? await scrapeCodein(htmlResponse)
+      : await scrapeMelange(htmlResponse, year)
+
+  return json
+}
+
+async function scrapeCodein(html) {
+  const $ = cheerio.load(html)
+  const mainSelector =
+    'section.home__winners > div.fixed-width > ul.no-style-list > li'
+
+  const resultObj = $(mainSelector).map((index, element) => ({
+    name: getListText($, element, 'div.home__winner-box-header > h3 > a')[0],
+    winners: getListText($, element, 'div:first-child > ul > li > strong'),
+    finalists: getListText($, element, 'div:last-child > ul > li'),
+  }))
+
+  return resultObj.get()
+}
+
+async function scrapeMelange(html, year) {
+  const $ = cheerio.load(html)
+
+  if (year < 2012) {
+    const mainSelector = 'div.main'
+
+    const resultObj = $(mainSelector).map((index, element) => ({
+      winners: getListText($, element, 'ul:first-of-type > li'),
+      organizations: getListText($, element, 'ul:last-child > li a'),
+    }))
+
+    return resultObj.get()
+  }
+
+  const mainSelector = 'div.org-winners-card'
+
+  const resultObj = $(mainSelector).map((index, element) => {
+    const orgInfo = {
+      name: getListText($, element, 'div.mdl-card__title > a')[0],
+      winners: getListText($, element, 'ul:first-of-type > li'),
+    }
+
+    if (year === 2014) {
+      orgInfo.finalists = getListText($, element, 'ul:last-child > li')
+    }
+
+    return orgInfo
+  })
+
+  return resultObj.get()
+}
+
+async function fetchStatistics(year, html) {
+  const $ = cheerio.load(html)
+  const statTable = $('table.wikitable.sortable > tbody > tr').slice(1)
+
+  const tableByYear = statTable
+    .filter(
+      (index, element) =>
+        removeInsideBrackets(getListText($, element, 'td')[0]) === String(year)
+    )
+    .first()
+
+  const tableData = getListText($, tableByYear, 'td')
+
+  const data = tableData.map(data => parseInt(data))
+
+  return {
+    number_of_organizations: data[1],
+    number_of_participants: data[2],
+    total_tasks_completed: data[3],
+  }
+}
+
+function getListText(html, element, selector) {
+  return html(element)
+    .find(selector)
+    .map((index, list) => html(list).text())
+    .get()
+}
+
+function removeInsideBrackets(str) {
+  return str.replace(/(\[.*?\]|\(.*?\)) */g, '')
+}
+
+module.exports = async () => {
+  const wikiResponse = await fetchText(CODEIN_WIKI_URL)
+
+  const fetchingPre2017 = PRE_URL.map(async url => {
+    const year = parseInt(url.split('/').pop())
+
+    return {
+      year,
+      statistics: await fetchStatistics(year, wikiResponse),
+      orgs: await fetchPre2017(year, url),
+    }
+  })
+
+  return await Promise.all(fetchingPre2017)
+}
--- a/tests/lib/pre2017.test.js
+++ b/tests/lib/pre2017.test.js
@@ -1,0 +1,17 @@
+/**
+ * Pre2017 Library Test
+ */
+
+const fs = require('fs')
+const pre2017 = require('../../lib/pre2017')
+
+describe('lib.pre2017', () => {
+  it('should have the correct contents', async () => {
+    const expectedData = JSON.parse(
+      fs.readFileSync(`${__dirname}/../__data__/pre2017.json`)
+    )
+    const pre2017Response = await pre2017()
+
+    expect(pre2017Response).toEqual(expectedData)
+  })
+})
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -7,6 +7,7 @@ const orgs = require('../out/data.json')
 const dates = require('../out/dates.json')
 const planet = require('../out/blog_planet.json')
 const manifest = require('../out/manifest.json')
+const pre2017 = require('../out/pre2017.json')
 
 const time = fs.statSync(`${__dirname}/../out/data.json`).mtime
 const datetime = new Date(time).toUTCString()
@@ -49,5 +50,14 @@ fs.writeFileSync(
   `${__dirname}/../out/planet/index.html`,
   Mustache.render(fs.readFileSync('templates/planet.html').toString(), {
     planet,
+  })
+)
+
+fs.mkdirSync(`${__dirname}/../out/pre2017`)
+
+fs.writeFileSync(
+  `${__dirname}/../out/pre2017/index.html`,
+  Mustache.render(fs.readFileSync('templates/pre2017.html').toString(), {
+    pre2017,
   })
 )
--- a/templates/pre2017.html
+++ b/templates/pre2017.html
@@ -1,0 +1,49 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Google Code In pre-2017 Finalists and winners</title>
+  </head>
+  <body>
+    <h1>Google Code In pre-2017 Finalists and winners</h1>
+    {{#pre2017}}
+      <div>
+        <h2>{{year}}</h2>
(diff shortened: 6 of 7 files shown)


lib/scrape.js
```diff
@@ -423,9 +434,88 @@ async function fetchDates() {
...
+  const agent = new https.Agent({
+    rejectUnauthorized: false,
+  })
+  const res = await fetch(url, { agent })
+  return await res.text()
+}
+
+async function fetchPre2017() {
+  const allData = []
+  for (let url of PRE_URL) {
```
blazeu: Loop should go outside the function.

Then we can test it easily.

lib/scrape.js
```diff
@@ -423,9 +434,88 @@ async function fetchDates() {
...
+    const year = url.split('/').pop()
+    console.log(`fetching GCI ${year}`)
+    const responseHTML = await fetchText(url)
```
blazeu: This can come from the arg so we can easily mock it.

blazeu (changes requested):
Put the function in its own file and call it from `lib/main.js`

Also create tests for the function.

See https://github.com/coala/gci-leaders/pull/95 for example.

tioirawan:
if I run ` npm test ` it give me an error `TypeError: Cannot read property 'close' of undefined`

lib/pre2017.js
```diff
@@ -0,0 +1,97 @@
...
+  })
+  const res = await fetch(url, {
+    agent,
+  })
+  return await res.text()
+}
+
+async function fetchPre2017(url) {
+  const year = url.split('/').pop()
+  const responseHTML = await fetchText(url)
```
blazeu: Fetching and scraping process should be separated.

Export the scraping function and test it.

blazeu (changes requested):
See RSS test for proper testing.

blazeu:
Might be a npm install corruption, run `npm install` again, you're using npm@5 right?

tioirawan:
now `npm test` work fine idk why, what should I test?

lib/main.js
```diff
@@ -2,6 +2,7 @@ const glob = require('glob')
 const fs = require('fs')
 
 const rss = require('./rss')
+const pre2017 = require('./pre2017')
```
wisn, line 5: Give a blank line after this `const` please.

lib/pre2017.js
```diff
@@ -0,0 +1,97 @@
...
+]
+
+async function fetchText(url) {
+  const agent = new https.Agent({
+    rejectUnauthorized: false,
+  })
+  const res = await fetch(url, {
+    agent,
+  })
+  return await res.text()
```
wisn: Give a blank line before `return` please.

lib/pre2017.js
```diff
@@ -0,0 +1,97 @@
...
+}
+
+async function fetchPre2017(url) {
+  const year = url.split('/').pop()
+  const responseHTML = await fetchText(url)
+  const html = cheerio.load(responseHTML)
+  let data = []
+  if (year == '2016' || year == '2015') {
```
wisn: Give a blank line before `if` block and use `===` instead of `==`.

lib/pre2017.js
```diff
@@ -0,0 +1,97 @@
...
+    const orgsList = html(`
```
wisn: Write

```javascript
const orgList = html(
  `...
   ...
   ...`
)

Instead?

File: lib/pre2017.js [hunk 8]
```diff
@@ -0,0 +1,97 @@
...
+          orgHTML,
+          'div.home__winner-box-content div:first-child ul li'
+        ),
+        finalists: getUlList(
+          orgHTML,
+          'div.home__winner-box-content div:last-child ul li'
+        ),
+      }
+    })
+  } else if (year == '2014' || year == '2013' || year == '2012') {
```
wisn: Use `===` instead. Or, read [in operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) if you're interested.

lib/pre2017.js
```diff
@@ -0,0 +1,97 @@
...
+    data = orgsList.toArray().map(org => {
+      const orgHTML = cheerio.load(org)
+      const info = {
+        name: orgHTML('div.mdl-card__title a').text(),
+        winners: getUlList(
+          orgHTML,
+          'div.mdl-card__supporting-text ul:first-of-type li'
+        ),
+      }
+      if (year == '2014')
```
wisn: Give a blank line before and after `if` block. Use `===` instead.

lib/pre2017.js
```diff
@@ -0,0 +1,97 @@
...
+        info.finalists = getUlList(
+          orgHTML,
+          'div.mdl-card__supporting-text ul:last-child li'
+        )
+      return info
+    })
+  } else if (year == '2011' || year == '2010') {
```
wisn: Use `===` instead.

lib/pre2017.js
```diff
@@ -0,0 +1,97 @@
...
+    const main = html('div.main')
+    data = main.toArray().map(org => {
+      const orgHTML = cheerio.load(org)
+      return {
+        winners: getUlList(orgHTML, 'ul:first-of-type li'),
+        organizations: getUlList(orgHTML, 'ul:last-child li a'),
+      }
+    })
+  }
+  return {
```
wisn: Give a blank before `return`.

wisn (changes requested):
No style for the web page yet?

lib/pre2017.js
tioirawan: `in` or `.includes()`?

lib/main.js
tioirawan: coala give a patch to remove line break

lib/pre2017.js
wisn: Whatever you feel better.

lib/main.js
wisn: Then patch it before push a commit.

lib/pre2017.js
```diff
@@ -0,0 +1,109 @@
...
+      return {
+        winners: getUlList(orgHTML, 'ul:first-of-type li'),
+        organizations: getUlList(orgHTML, 'ul:last-child li a'),
+      }
+    })
+  }
+
+  return {
+    year,
+    generated: date,
```
blazeu: I don't think this is needed.

tests/lib/pre2017.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+const fs = require('fs')
+const pre2017 = require('../../lib/pre2017')
+const mockdate = require('mockdate')
+
+describe('lib.pre2017', () => {
+  let pre2017Response;
+
+  beforeEach(async () => {
+    mockdate.set('2017-12-31T04:14:05.214Z')
```
blazeu: Not necessary.

lib/pre2017.js
```diff
@@ -0,0 +1,109 @@
...
+}
+
+async function fetchPre2017(url, date) {
+  const year = url.split('/').pop()
+  const responseHTML = await fetchText(url)
+
+  return await scrape(responseHTML, year, date)
+}
+
+async function scrape(responseHTML, year, date) {
```
blazeu: 1. export this function
2. call it from the test
3. provide 2 "fake" data; html & json result from the scrapping process
3. assert the resulting HTML matches json result

blazeu (changes requested):

lib/pre2017.js
blazeu: Or you could mock the `fetch` so that it will return the html without network access.
and test the function in `module.exports`

That will test the whole file, but it is too complex.

lib/pre2017.js
```diff
@@ -0,0 +1,109 @@
...
+const https = require('https')
+
+const PRE_URL = [
+  'https://codein.withgoogle.com/archive/2016',
+  'https://codein.withgoogle.com/archive/2015',
+  'https://www.google-melange.com/archive/gci/2014', // Ignore InvalidLinkBear
+  'https://www.google-melange.com/archive/gci/2013', // Ignore InvalidLinkBear
+  'https://www.google-melange.com/archive/gci/2012', // Ignore InvalidLinkBear
+  'https://www.google-melange.com/archive/gci/2011', // Ignore InvalidLinkBear
+  'https://www.google-melange.com/archive/gci/2010', // Ignore InvalidLinkBear
```
blazeu: Modify `.coafile` to use `link_ignore_list`

ref https://github.com/coala/projects/blob/master/.coafile#L6

lib/pre2017.js
tioirawan: so I need to compare scraped data from url and scraped data from 'mocked' html?

blazeu: This function takes html string, right? Use that downloaded html to call the function.

Roughly the test will be like this
```js
const res = await scrape(html, year)
expect(res).toEqual(expectedResult)
```

where the `expectedResult` contains the expected result of the function call with that corresponding html.

.coafile
```diff
@@ -1,6 +1,7 @@
 [all]
 files = **.js, **.json, **.css, **.yml, **.html, .eslintrc
 ignore = node_modules/**, out/**, tests/__data__/**, package-lock.json
+link_ignore_list = https://www.google-melange.com/**
```
blazeu: Put it in `all.links`

tests/lib/pre2017.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+describe('lib.pre2017', () => {
+  const basedir = `${__dirname}/../__data__/pre2017html`
+  let pre2017Response
+
+  beforeEach(async () => {
+    pre2017Response = await pre2017.get()
+  })
+
+  test('the json results for correct data', async () => {
```
blazeu: Use `it('should..`

lib/pre2017.js
```diff
@@ -0,0 +1,115 @@
...
+  return html(selector)
+    .toArray()
+    .map(list =>
+      html(list)
+        .text()
+        .replace(/[\n\t]/g, '')
+    )
+}
+
+async function get() {
```
blazeu: Let's just back off to the beginning; you seem to not understand

export this function only, like before 

and see my another review on the test.

tests/lib/pre2017.test.js
```diff
@@ -0,0 +1,42 @@
...
+    const fetchingExpectedData = pre2017.PRE_URL.map(async url => {
+      const year = url.split('/').pop()
+      return await pre2017.scrape(
+        fs.readFileSync(`${basedir}/${year}.html`).toString(),
```
blazeu: It is **really** *really* weird to have the html but still fetching it via `pre2017.get()` and `get()` is calling `scrape` again internally, so what's the point of this test?

Let's go with this approach and see if you understand.
The function job is to generate a javascript object,
we need to make sure that the object is correct,
for that, you have to have the actual object first,
run the function once (outside the test; manually running it by yourself),
store the return value somewhere (json file?)

After that, you create a test, calling `pre2017`,
and test that it returns the same object that you have before.

blazeu (changes requested):

tests/lib/pre2017.test.js
tioirawan: It's look like what i've done before

blazeu: Your previous one is only testing if the json file exist?

We need it to be correct, not just "existed"; pardon me if you were already asserting it to be correct before.

Indmind:
```js

Comment (tioirawan) [reply tests/lib/pre2017.test.js]: Like this?
```js
/**
 * Pre2017 Library Test
 */

const fs = require('fs')
const pre2017 = require('../../lib/pre2017')

describe('lib.pre2017', () => {

  it('should have the correct contents',async () => {
    const expectedData = JSON.parse(
      fs.readFileSync(`${__dirname}/../__data__/pre2017.json`)
    )
    const pre2017Response = await pre2017()
    const pre2017json = pre2017Response.map(data => data.json)

    expect(pre2017json).toEqual(expectedData)
  })

})
```

blazeu: Yep

lib/pre2017.js
```diff
@@ -0,0 +1,106 @@
...
+
+async function fetchPre2017(url) {
+  const year = url.split('/').pop()
+  const html = await fetchText(url)
+  const json = await scrape(html, year)
+
+  return { html, json }
+}
+
+async function scrape(responseHTML, year) {
```
jayvdb: split into two functions, one for codein.withgoogle.com and the other for melange.

they are wildly different scrapers, as the server uses different software, with rather different capabilities.

the codein scraper (2015+) should work on 2017, after 2017 codein is finished.

both scrapers will be improved more after this task is completed (e.g. extracting the tasks each person completed)

lib/pre2017.js
```diff
@@ -0,0 +1,106 @@
...
+  })
+  const res = await fetch(url, {
+    agent,
+  })
+
+  return await res.text()
+}
+
+async function fetchPre2017(url) {
+  const year = url.split('/').pop()
```
jayvdb: year should be an integer.

lib/main.js
```diff
@@ -17,4 +18,11 @@ const rss = require('./rss')
   const rssResult = rss(data)
   fs.writeFileSync(`${__dirname}/../out/feed_items.json`, rssResult.feed_items)
   fs.writeFileSync(`${__dirname}/../out/feed.xml`, rssResult.feed_xml)
+
+  const pre2017Result = await pre2017()
+  const pre2017JSON = pre2017Result.map(data => data.json)
+  fs.writeFileSync(
+    `${__dirname}/../out/pre2017.json`,
+    JSON.stringify(pre2017JSON)
```
jayvdb: pretty print the JSON like the other emitted JSON files
(and the JSON should have an EOL at EOF)

jayvdb (dismissed):

lib/main.js
tioirawan: I can't find JSON.strigify option to add OEL

tioirawan:
I don't know what else to do, so I add statistic

tests/lib/pre2017.test.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Pre2017 Library Test
+ */
+
+const fs = require('fs')
+const pre2017 = require('../../lib/pre2017')
+
+describe('lib.pre2017', () => {
+
+  it('should have the correct contents',async () => {
```
wisn: Space after comma lol

tests/lib/pre2017.test.js
```diff
@@ -0,0 +1,19 @@
...
+    const expectedData = JSON.parse(
+      fs.readFileSync(`${__dirname}/../__data__/pre2017.json`)
+    )
+    const pre2017Response = await pre2017()
+
+    expect(pre2017Response).toEqual(expectedData)
```
wisn, line 15: Have you tried `toContainEqual` instead? See [here](https://facebook.github.io/jest/docs/en/expect.html#tocontainequalitem).

lib/pre2017.js
```diff
@@ -0,0 +1,130 @@
...
+  const mainSelector = 'div.org-winners-card'
+
+  const resultObj = $(mainSelector).map((index, element) => {
+    const orgInfo = {
+      name: getListText($, element, 'div.mdl-card__title > a')[0],
+      winners: getListText($, element, 'ul:first-of-type > li'),
+    }
+
+    if (year === 2014)
+      orgInfo.finalists = getListText($, element, 'ul:last-child > li')
```
wisn: If you're using a new line after `if` statement, use curly brackets.

```javascript
// good
if (year === 2014) ...

// bad
if (year === 2014)
  ...

// good
if (year === 2014) {
  ...
}
```

Why?

- See [AirBnB JS Guide](https://github.com/airbnb/javascript#blocks--braces).
- Read more about [ASI](https://tc39.github.io/ecma262/#sec-automatic-semicolon-insertion).

wisn (changes requested):

tests/lib/pre2017.test.js
tioirawan: if I test using `toContainEqual` the test fail

blazeu (approved):

blazeu:
ack https://github.com/coala/gci-leaders/pull/118/commits/105fcde609d78d1ce67a203619251fb980f92609

Issue #125, opened 2018-01-02
Generate language list dynamically


blazeu:
https://github.com/coala/gci-leaders/blob/9a170ecf6e39813f0a722fee6a5ba8378b9af3cb/src/js/locale.js#L29-L34

It needs to be generated dynamically, from the locale files that exists in [static/i18n](https://github.com/coala/gci-leaders/tree/master/static/i18n) to support https://github.com/coala/gci-leaders/issues/124.

We can generate a JSON file and `import` that file to replace the JS object in locale.js

The language name can come from a npm package, but do not include that package in the client side js as to not bloat it.

mooncos:
I have an issue with this. Webpack bundles all js files before I can add my own with the translations hash list.

blazeu:
You should take advantage of the bundling, use `import` to import that hash/object json file.

The file should be generated before the bundling happens, obviously.

PR #136, opened 2018-01-14
langParser.js: Generate language list dynamically

mooncos:
This adds ``langParser.js`` to generate the available
language list automatically from available translation
files. This also adds a npm script for generating before
webpack bundles. Modifies .gitignore so that generated
``languages.js`` file is ignored. Adds tests. Renames
``nb_NO.json`` to ``nb-NO.json`` to comply with RFC
standards.

Closes: #125


lib/langParser.js
```diff
@@ -0,0 +1,20 @@
+const fs = require('fs')
+const iso639 = require('iso-639-1')
+
+fs.readdir(`${__dirname}/../static/i18n`, function(err, items) {
```
andrewda: ES6 syntax: `(err, items) => {`

lib/langParser.js
```diff
@@ -0,0 +1,20 @@
...
+  var avaliableLanguages = {
+    Languages: {},
+  }
+  for (var i = 0; i < items.length; i++) {
```
andrewda: `items.forEach(...)` is significantly cleaner in this case

lib/langParser.js
```diff
@@ -0,0 +1,20 @@
+const fs = require('fs')
+const iso639 = require('iso-639-1')
+
+fs.readdir(`${__dirname}/../static/i18n`, function(err, items) {
+  var avaliableLanguages = {
+    Languages: {},
+  }
```
andrewda: Does this need to have a nested object? Would be much cleaner to just use `availableLanguages = {}`

lib/langParser.js
```diff
@@ -0,0 +1,20 @@
+const fs = require('fs')
+const iso639 = require('iso-639-1')
+
+fs.readdir(`${__dirname}/../static/i18n`, function(err, items) {
+  var avaliableLanguages = {
```
andrewda: Typo - should be `availableLanguages`.

Also, use ES6's `const` instead of `var`.

lib/langParser.js
```diff
@@ -0,0 +1,20 @@
...
+    Languages: {},
+  }
+  for (var i = 0; i < items.length; i++) {
+    var langName = items[i].substring(0, items[i].indexOf('.'))
+    var normalizedName = langName.split('-')[0]
+    var nativeName = iso639.getNativeName(normalizedName)
```
andrewda: All of these can be `const` instead of `var`.

lib/langParser.js
```diff
@@ -0,0 +1,20 @@
...
+    avaliableLanguages['Languages'][nativeName] = langName
+  }
+  fs.writeFileSync(
+    `${__dirname}/../src/js/languages.js`,
+    'export default ' +
+      JSON.stringify(avaliableLanguages) +
+      ' // eslint-disable-line'
```
andrewda: This is pretty sketch. Can we write directly to a `languages.json` JSON file instead?

andrewda (changes requested):
Also, the rename from nb_NO to nb-NO seems unrelated to this PR.

lib/langParser.js
mooncos: It needs to be so that we can import all nested objects at once.

lib/langParser.js
mooncos: I've tried with no success...

mooncos:
@andrewda it needs to be changed to nb-NO as per [ISO and RFC standards](https://tools.ietf.org/html/rfc5646#section-2.1) compliance for the npm package

lib/langParser.js
mooncos: Is it possible to import a JSON file in ES6? I can't seem to achieve it.

jayvdb:
We cant rename `static/i18n/nb_NO.json` → `static/i18n/nb-NO.json` .  These filenames are provided by Weblate, so we must use them otherwise they think the translation is missing.

mooncos:
Ok.

lib/langParser.js
Indmind: https://blog.jonathanargentiero.com/how-to-load-a-json-content-inside-a-js-module-with-webpack/

lib/langParser.js
```diff
@@ -0,0 +1,20 @@
...
+  }
+  items.forEach(function(value) {
+    const langName = value.substring(0, value.indexOf('.'))
+    const normalizedName = langName.split('-')[0]
+    const nativeName = iso639.getNativeName(normalizedName)
+    availableLanguages['Languages'][nativeName] = langName
+  })
+  fs.writeFileSync(
+    `${__dirname}/../src/js/languages.js`,
+    'export default ' +
```
blazeu: A `.json` file saves the trouble of having to `export` things.

blazeu (changes requested):

blazeu:
There's a webpack plugin to generate json file
https://www.npmjs.com/package/generate-json-webpack-plugin

Using that will prevent us from having another js script be added to build script.

lib/langParser.js
mooncos: And how can I import a JSON file?

blazeu: `import name from 'file.json'`

mooncos: I’ve tried without success

blazeu: try `require()` ?

mooncos:
Ok. Should I write all logic within `webpack.config.js`?

blazeu:
Not really, you can import/require `langParser.js` in `webpack.config.js`

mooncos:
@blazeu fixed that. but where is the generated JSON located?

src/js/languages.json
```diff
@@ -0,0 +1 @@
+{"English":"en","Español":"es","Norsk bokmål":"nb_NO","język polski":"pl"}
```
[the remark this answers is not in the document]
jayvdb: Can you make this pretty printed.

Also I guess it should be in `static/` , but I am not 100% sure about that.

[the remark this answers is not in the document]
andrewda: Or `.gitignore` it because it's generated each time.

margobra8:
I fixed that but I now have another problem: The webpack reads the `import` statements _before_ the `languages.json` file is generated.

src/js/languages.json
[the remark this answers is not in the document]
mooncos: I fixed that but I now have another problem: webpack reads the `import` statements _before_ the `languages.json` file is generated. I don’t know how to fix this.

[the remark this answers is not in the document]
blazeu: https://github.com/elliottsj/generate-json-webpack-plugin/issues/5

Yup it is not possible to import the generated json, might have to use the previous method or search for another webpack plugin.

jayvdb (approved):

jayvdb:
ack https://github.com/coala/gci-leaders/pull/136/commits/6948c27a291c68bd3e0854537bf35ada3d5253a6

jayvdb:
needs a rebase, and a recheck by @blazeu , or @wisn

webpack.config.js
```diff
@@ -104,6 +106,7 @@ module.exports = {
     ]),
     new ExtractTextPlugin(`[name]${hash}.css`),
     new ManifestPlugin(),
+    new GenerateJsonPlugin('../src/js/languages.json', generatedLangList),
```
wisn, line 109: In case you want to prettify the output.

```javascript
GenerateJsonPlugin(<file>, <json>, [replacer], [space])
```

Fill out that `space` argument.

wisn (approved):

webpack.config.js
mooncos: Ok

src/js/languages.json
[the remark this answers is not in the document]
blazeu: >webpack reads the import statements before the languages.json file is generated.

When new language is added, the compiled js will still import the file that has the old language list. That is undesirable.

PR #139, opened 2018-01-16
Update deadline

daniellimws:
Currently the progress is exceeding 100% as it is looking at the wrong deadline.

Update the deadline used to report the actual time left.

### DIFF
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -14,7 +14,7 @@ const datetime = new Date(time).toUTCString()
 const rootURL = process.env.URL
 
 const competitionOpen = new Date(dates.competition_open_starts)
-const noClaims = new Date(dates.competition_open_ends)
+const noClaims = new Date(dates.grace_period_ends)
 
 const noLeader = []
 const withLeader = orgs.filter(org => {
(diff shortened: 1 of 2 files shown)


mkljczk (approved):

blazeu:
This will be negative numbers again when the competition really ends. Not really a "fix".
Editing the locale file is also not good, since you have to update all of the language.

I think we have a `timeDifference` function for handling past time. 
Go edit this instead https://github.com/coala/gci-leaders/blob/master/src/js/app.js#L36

Yeah, having 2 separate function for generating relative time is _stupid_, I know.
I'd be really happy if you refactor that into a single function, but not mandatory.

daniellimws:
Sure, I'll work on that tomorrow, though may be a bit lazy because 2 weeks later this website will be kinda obsolete.

jayvdb:
unack https://github.com/coala/gci-leaders/pull/139/commits/2823ef61aaa67a94650f4798b495e290c1088e13

Makman2:
Please don't close PRs just because they weren't acked, you can always audit a PR.

PR #148, opened 2018-04-16
scrape.js: Use GraphQL API

### LINKED ISSUE #111: Use GraphQL API

li-boxuan:
Use GitHub GraphQL API instead of REST API

closes https://github.com/coala/gci-leaders/issues/111

### DIFF
--- a/lib/queries/github_search_org.graphql
+++ b/lib/queries/github_search_org.graphql
@@ -1,0 +1,9 @@
+query($query: String!) {
+  search(type: USER, query: $query, first: 1) {
+    nodes {
+      ...on Organization {
+          login
+      }
+    }
+  }
+}
--- a/lib/queries/github_user_info.graphql
+++ b/lib/queries/github_user_info.graphql
@@ -1,0 +1,6 @@
+query($user: String!) {
+  user(login: $user) {
+    login
+    updatedAt
+  }
+}
--- a/lib/scrape.js
+++ b/lib/scrape.js
@@ -8,7 +8,11 @@ const validUsername = require('valid-github-username')
 const wdk = require('wikidata-sdk')
 const cheerio = require('cheerio')
 
-const { GITHUB_REPO_INFO_QUERY } = require('./queries')
+const {
+  GITHUB_REPO_INFO_QUERY,
+  GITHUB_SEARCH_ORG_QUERY,
+  GITHUB_USER_INFO_QUERY,
+} = require('./queries')
 const { getLatestCommitMessage } = require('./utils')
 
 const GH_BASE = 'https://github.com'
@@ -137,6 +141,7 @@ async function fetchRepositoryInfo(org) {
     ;({ data, errors } = await client.query(GITHUB_REPO_INFO_QUERY, { org }))
   } catch (error) {
     console.warn(`GitHub query for org ${org} fails, error: ${error}`)
+    return []
   }
 
   if (data && data.organization) {
@@ -209,12 +214,34 @@ async function checkGitHubUserExists(user) {
 }
 
 async function searchGitHubOrgs(query) {
+  let results = []
+  // use REST API to fetch
   const res = await fetch(
-    `${GH_API_BASE}/search/users?q=${query}%20type:org`,
+    `${GH_API_BASE}/search/users?q=${query}`,
     GH_API_OPTIONS
   )
+  console.log('query=', `${GH_API_BASE}/search/users?q=${query}`)
   const { items } = await res.json()
-  return items || []
+  results = results.concat(items)
+  // use GraphQL API to fetch
+  let data, errors
+  try {
+    ;({ data, errors } = await client.query(GITHUB_SEARCH_ORG_QUERY, { query }))
+  } catch (error) {
+    console.warn(`GitHub query ${query} fails, error: ${error}`)
+  }
+
+  if (data && data.search && data.search.nodes) {
+    results = results.concat(data.search.nodes)
+  } else {
+    const errorMessage =
+      errors && errors.length ? errors[0].message : 'unknown error'
+    console.warn(
+      `Cannot query ${query} from GitHub, error message: ${errorMessage}`
+    )
+  }
+  console.log('results', results)
+  return unique(results)
 }
 
 async function getGitHubUserHistory(user, from, to) {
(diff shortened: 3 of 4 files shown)


li-boxuan:
I add [L29](https://github.com/coala/gci-leaders/pull/148/files#diff-9676cbcb5938ab810f7fa82a528abc4eR29) and [L114](https://github.com/coala/gci-leaders/pull/148/files#diff-e7f5b3f014790acdf91e958a37dc5f2cR114) to make CI pass. (CI was failing on HEAD). Not sure if I should keep them in this PR (by adding commits?) or file a new issue to solve that first.

jayvdb:
I notice that JBoss has two linked on https://gci-leaders.netlify.com/ , but none on https://deploy-preview-148--gci-leaders.netlify.com/ .

lib/scrape.js
```diff
@@ -9,17 +9,16 @@ const wdk = require('wikidata-sdk')
 const cheerio = require('cheerio')
 
 const { GITHUB_REPO_INFO_QUERY } = require('./queries')
+const { GITHUB_SEARCH_ORG_QUERY } = require('./queries')
+const { GITHUB_USER_INFO_QUERY } = require('./queries')
```
blazeu: Just do multiple variable on a single destructuring line { a, b, c }

blazeu (dismissed):

li-boxuan:
@jayvdb I just investigated that problem (JBoss has no link).

This is not due to this PR; I use the previous HEAD to build and also found no link on JBoss part.

This is because of the name of their gitter room: this link https://gitter.im/jboss-outreach/gci (See [#258](https://github.com/coala/gci-leaders/blob/master/lib/scrape.js#L258), which makes use of that link to extract their org name).

However, their org name should be `JBossOutreach` instead of `jboss-outreach`. That's why two links are missing. Btw, https://gitter.im/jboss-outreach/gci would automatically jump into https://gitter.im/JBossOutreach/gci, which is indeed what we want.

I added the following code around line [#272](https://github.com/coala/gci-leaders/blob/master/lib/scrape.js#L272) and ran it on my machine: JBoss would now again have two links.
```javascript
if (orgFromWebsites === 'jboss-outreach') {
    return 'JBossOutreach'
}
```

Do you have some advice what I should do to fix this problem? Visit every link and see if it would jump to some other link? Or add ad-hoc support for `JBossOutreach`?

lib/scrape.js
li-boxuan: Thx, done.

jayvdb:
fwiw, a part of the problem was that https://github.com/coala/gci-leaders/commit/92a68a288c490adc2c2d2ef2854dba982d41db21 cleared out all of the data in the deploy.

We dont need a special case for JBoss.  The link in GCI is old, but a head request will give us the new URL.

That is needed for any github/gitter URL which changes.
That should be a new PR which just fixes master.

lib/rss.js
```diff
@@ -25,8 +25,11 @@ module.exports = ({
   feed_items: current,
   data_updated: dataUpdated,
 }) => {
-  const oldOrgs = createOrgHash(oldData)
   const newOrgs = createOrgHash(newData)
+  let oldOrgs = newOrgs
```
bekicot: Does it means the oldOrgs is no longer in use?

Update
Nevermind, was confused with the other pull. Looks good.

jayvdb:
It would be good to finish this up, as I suspect we will need this in order to do https://github.com/coala/gci-leaders/issues/165 , and possible also https://github.com/coala/gci-leaders/issues/164

lib/queries/github_user_info.graphql
```diff
@@ -0,0 +1,6 @@
+query ($user: String!) {
+    user(login: $user) {
```
jayvdb: our other graphql file uses two space indent.

if you think four is better, also create an issue to re-indent the existing file, and we'll make it a newcomer task

lib/scrape.js
```diff
@@ -53,12 +54,6 @@ const CHAT_IMAGES = {
   OTHER: 'images/chat.png',
 }
 
-const GH_API_OPTIONS = {
-  headers: process.env.GITHUB_TOKEN
-    ? { Authorization: `token ${process.env.GITHUB_TOKEN}` }
```
jayvdb: keep this, and keep GH_API_BASE, as I am sure we will want to access some features are not available via Graphql, and we dont want people to need to re-invent this, and possibly pick a new token variable name, etc.

Add a note that `GH_API_OPTIONS` is unused.

jayvdb:
Lots of "Cannot fetch org x from GitHub, error message: unknown error" in the log https://travis-ci.org/coala/gci-leaders/builds/447609704

lib/queries/github_user_info.graphql
li-boxuan: I should have used two space indent. Blame my editor :P

lib/queries/github_user_info.graphql
```diff
@@ -0,0 +1,6 @@
+query ($user: String!) {
```
andrewda: Remove space before `(`
(or add a space on the other ones, I'm not 100% sure which is preferred in GraphQL, but I think I've seen no space used more frequently)

```suggestion
query($user: String!) {
```

lib/queries/github_search_org.graphql
```diff
@@ -0,0 +1,9 @@
+query($query: String!) {
+    search(type:USER, query:$query, first:1) {
```
andrewda: Add space after colons:

```suggestion
    search(type: USER, query: $query, first: 1) {
```

andrewda (changes requested):

li-boxuan:
> Lots of "Cannot fetch org x from GitHub, error message: unknown error" in the log

Weird, it works fine on my local machine. The log should be sth as follows:
`Cannot fetch user xxx from GitHub, error message: Could not resolve to a User with the login of 'xxx'.`

andrewda:
IIRC the number of unauthenticated GraphQL requests you get is lower than normal API hits. Since this is a PR, I think Travis removes the `GH_TOKEN` env variable for security reasons, so all the GraphQL requests are made unauthenticated. Might be why it is unable to fetch users after a certain point?

jayvdb (changes requested):
also keep the old org/user lookup, and use them as a fallback after the graphsql has done its magic

lib/scrape.js
```diff
@@ -209,12 +215,24 @@ async function checkGitHubUserExists(user) {
...
 async function searchGitHubOrgs(query) {
-  const res = await fetch(
-    `${GH_API_BASE}/search/users?q=${query}%20type:org`,
-    GH_API_OPTIONS
-  )
-  const { items } = await res.json()
-  return items || []
+  let data, errors
+  try {
+    ;({ data, errors } = await client.query(GITHUB_SEARCH_ORG_QUERY, { query }))
```
bekicot, line 229: Interesting, what is the semicolon for?

li-boxuan: Not sure, but `eslint` will be unhappy with that ;)

bekicot: I guess it is because you put parentheses in
```
;({ data, errors } = await client.query(GITHUB_SEARCH_ORG_QUERY, { query }))
```

Another question, why is the parentheses used?

li-boxuan: Yes, this is because I put parentheses.
I have to use parentheses otherwise there would be a parsing error.

jayvdb (approved):

jayvdb:
ack https://github.com/coala/gci-leaders/pull/148/commits/8da2685684732dd8e6fff5c59da2c8267611a906

jayvdb:
@gitmate-bot ff

PR #152, opened 2018-06-16
rss.js: Fix TypeError for undefined oldData

### LINKED ISSUE #150: Error: Cannot find module '../out/pre2017.json'
See error at https://github.com/coala/gci-leaders/pull/149#issuecomment-386820394

hizkifw:
Fixes #150


bekicot:
This is could be the fix. I would also recommend updating the corresponding test.

My personal preference is to log a warning whenever the oldData is null or undefined. As it should not be empty.

cc @andrewda && @blazeu for that. It is maybe not a good idea to log a warning.

lib/rss.js
```diff
@@ -25,8 +25,11 @@ module.exports = ({
   feed_items: current,
   data_updated: dataUpdated,
 }) => {
-  const oldOrgs = createOrgHash(oldData)
   const newOrgs = createOrgHash(newData)
+  let oldOrgs = newOrgs
+  if (oldData) {
+    oldOrgs = createOrgHash(oldData)
+  }
```
wisn, line 32: Actually, we could do this:

```javascript
const oldOrgs = createOrgHash(oldData) || createOrgHash(newdata)
```

Please not that the return should be falsy object such as false boolean, undefined, null, or zero.

wisn (commented):
Well, everything is good. My review was trivial. It's for elegancy only.

As for the tests, it seems you are use string. So, please take a look at [Jest cheatsheet](https://devhints.io/jest) and follow it. Test to string use `toMatch` instead of `toEqual`. It is trivial so you may just ignore this.

lib/rss.js
hizkifw: We can't do that because if we pass an undefined oldData to createOrgHash, it would throw an exception. We could wrap the function inside a try/catch block, but I'm not sure if that's better than using an if statement.

wisn: You may use `createOrgHash(oldData || 'something')`. As I said, it's trivial.

tests/__data__/feed_items_no_old.json
```diff
@@ -0,0 +1 @@
...ted Tasks Updated for Liquid Galaxy project","description":"Liquid Galaxy project changed completed_task_instance_count from 153 to 159.","date":"2017-12-24T03:04:19.533Z","custom_elements":[{"org:name":"Liquid Galaxy project"},{"org:slug":"liquid-galaxy-project"},{"org:id":5067404587565056},{"property:path":"completed_task_instance_count"},{"property:type":"change"},{"property:old":153},{"property:new":159}]},{"title":"Completed Tasks Updated for Drupal","description":"Drupal changed completed_task_instance_count from 136 to 142.","date":"2017-12-24T03:04:19.533Z","custom_elements":[{"org:name":"Drupal"},{"org:slug":"drupal"},{"org:id":5636094664114176},{"property:path":"completed_task_instance_count"},{"property:type":"change"},{"property:old":136},{"property:new":142}]},{"title":"Completed Tasks Updated for Apertium","description":"Apertium changed completed_task_instance_count from 70 to 72.","date":"2017-12-24T03:04:19.533Z","custom_elements":[{"org:name":"Apertium"},{"org:slug":"apertium"},{"org:id":6541403171586048},{"property:path":"completed_task_instance_count"},{"property:type":"change"},{"property:old":70},{"property:new":72}]}],"lastUpdated":"2017-12-24T03:04:19.533Z"}
```
jayvdb: reformat as 2 space intended JSON, and trim it to a *minimal* test case.

jayvdb (changes requested):

lib/rss.js
blazeu: We could use ternary if we really want it short

```js
const oldOrgs = oldData ? createOrgHash(oldData) : newOrgs
```

but, maybe, it's clearer if we use explicit conditional.

blazeu (approved):

tests/__data__/feed_alt.xml
```diff
@@ -0,0 +1 @@
+<?xml version="1.0" encoding="UTF-8"?><rss xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:content="http://purl.org/rss/1.0/modules/content/" xmlns:atom="http://www.w3.org/2005/Atom" version="2.0" xmlns:org="https://g.co/gci" xmlns:property="https://g.co/gci"><channel><title><![CDATA[Google Code-in Leaders]]></title><description><![CDATA[A feed for Google Code-in updates]]></description><link>https://gci-leaders.netlify.com</link><generator>RSS for Node</generator><lastBuildDate>Sun, 24 Dec 2017 03:04:35 GMT</lastBuildDate><atom:link href="https://gci-leaders.netlify.com/feed.xml" rel="self" type="application/rss+xml"/><pubDate>Sun, 24 Dec 2017 03:04:35 GMT</pubDate></channel></rss>
```
blazeu, line 1: Missing newline

wisn (approved):

bekicot (approved):

tests/__data__/feed_alt.xml
hizkifw: The feed generated by `rss.js` doesn't contain a trailing `\n`, and so adding it in this file would fail the test. Would it be better to add a newline to the test of leave the file as-is?

blazeu: ok leave as is

jayvdb:
@gitmate-bot ff

jayvdb (approved):

PR #171, opened 2018-11-28
Indonesian Translation

drepram:
Is there a Google Code-in task for this? I really hope that it is.


static/i18n/id.json
```diff
@@ -0,0 +1,21 @@
+{
+  "page-title": "Pemimpin Sementara Google Code-in 2018",
```
andrewda: Indentation should be 4 spaces

andrewda (changes requested):
I think you've got some wacky characters before each line that aren't spaces 😄 Make sure each line starts with 4 space characters!

drepram:
Changed it to 4 spaces.

Then again, my questions still stands, is there a task for this? I can't seem to find it.

static/i18n/id.json
drepram: Fixed!

PR #157, opened 2018-10-08
lib/scrape.js: Add year variable

utkarsh2102:
This fixes the year by creating a GCI_YEAR variable.

Fixes https://github.com/coala/gci-leaders/issues/153

PR #176, opened 2019-12-04
Add Strings Replacing GCI 2018  with GCI 2019

MatejMecka:
Website says Google Code-in 2018 Leaders but Google Code-in 2018 ended and now it's Google Code-in 2019

### DIFF
--- a/static/i18n/en.json
+++ b/static/i18n/en.json
@@ -1,5 +1,5 @@
 {
-    "page-title": "Google Code-in 2018 Current Leaders",
+    "page-title": "Google Code-in 2019 Current Leaders",
     "last-updated": "Last updated",
     "order-info": "The leading participants for each organization (and their GitHub accounts, if applicable) are listed randomly.",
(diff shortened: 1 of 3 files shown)


CLiu13 (approved):
Commit message is a bit ambiguous (you aren't adding the strings, but rather you are just updating them) - but aside from that, LGTM. 👍

We should probably figure out a way to have the year auto-update (see https://github.com/coala/gci-leaders/pull/157#pullrequestreview-169048770), but I suppose that's another task for another time.

kx-chen (approved):
LGTM as well, I agree with Charlie's comments about making the year dynamic. It's definitely another issue for later. 

Thanks Matej! :)

yukiisbored (approved):

yukiisbored:
Will ff merge locally since GitMate is no more. RIP GitMate.

yukiisbored:
Thank you for the patch.

